PR #4, opened 2026-03-11
fix(memory): honor explicit window limit over env fallback

kurozee:
## Summary

  - Explicit memory window limits were being reduced by ambient `LOONGCLAW_SLIDING_WINDOW`.
  - This change makes caller-provided limits win, while keeping env/default behavior only as fallback when no limit is supplied.
  - The hard cap at `128` remains unchanged.
  - Regression tests were added for explicit-limit precedence, default fallback, and cap behavior.

  ## Scope

  - [x] Small and focused
  - [x] No unrelated refactors

  ## Risk Track

  - [x] Track A (routine/low-risk)
  - [ ] Track B (higher-risk/policy-impacting)

  ## Validation

  - [ ] Additional scenario/benchmark checks (if applicable)

  Additional validation run:

  ## Linked Issues
Closes #

### DIFF
--- a/crates/app/src/memory/mod.rs
+++ b/crates/app/src/memory/mod.rs
@@ -235,8 +235,13 @@ mod tests {
         };
 
         for idx in 0..3 {
-            append_turn_direct("window-limit-session", "user", &format!("turn-{idx}"), &config)
-                .expect("append_turn_direct should succeed");
+            append_turn_direct(
+                "window-limit-session",
+                "user",
+                &format!("turn-{idx}"),
+                &config,
+            )
+            .expect("append_turn_direct should succeed");
         }
 
         std::env::set_var("LOONGCLAW_SLIDING_WINDOW", "1");
@@ -270,8 +275,13 @@ mod tests {
         };
 
         for idx in 0..130 {
-            append_turn_direct("window-default-session", "user", &format!("turn-{idx}"), &config)
-                .expect("append_turn_direct should succeed");
+            append_turn_direct(
+                "window-default-session",
+                "user",
+                &format!("turn-{idx}"),
+                &config,
+            )
+            .expect("append_turn_direct should succeed");
         }
 
         std::env::set_var("LOONGCLAW_SLIDING_WINDOW", "3");
--- a/crates/app/src/memory/mod.rs
+++ b/crates/app/src/memory/mod.rs
@@ -216,4 +216,93 @@ mod tests {
         let _ = fs::remove_file(&db_path);
         let _ = fs::remove_dir(&tmp);
     }
+
+    #[cfg(feature = "memory-sqlite")]
+    #[test]
+    fn explicit_window_limit_ignores_env_default() {
+        use std::fs;
+
+        let tmp = std::env::temp_dir().join(format!(
+            "loongclaw-test-memory-window-limit-{}",
+            std::process::id()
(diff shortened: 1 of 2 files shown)


crates/app/src/memory/sqlite.rs
```diff
@@ -80,13 +80,11 @@ pub(super) fn load_window(
         .map(str::trim)
         .filter(|value| !value.is_empty())
         .ok_or_else(|| "memory.window requires payload.session_id".to_owned())?;
-    let requested_limit = payload
+    let window_limit = payload
```
gh-xj, line 83: This fixes the explicit-limit path, but it also regresses the fallback path: `default_window_size()` is no longer clamped to 128. Since `LOONGCLAW_SLIDING_WINDOW` comes from config and `memory.sliding_window` has no upper-bound validation, `memory.window` requests without an explicit `limit` can now exceed the previous hard cap. Please choose the source first, then apply `.clamp(1, 128)` once.

crates/app/src/memory/mod.rs
```diff
@@ -216,4 +216,103 @@ mod tests {
...
+            append_turn_direct(
+                "window-limit-session",
+                "user",
+                &format!("turn-{idx}"),
+                &config,
+            )
+            .expect("append_turn_direct should succeed");
+        }
+
+        std::env::set_var("LOONGCLAW_SLIDING_WINDOW", "1");
```
gh-xj, line 247: This test mutates the process-global `LOONGCLAW_SLIDING_WINDOW` env var without restoring the previous value or serializing access. The same issue appears again later in this file. That makes the suite flaky under parallel test execution. Please guard env mutation and restore the original value, or merge these cases into one deterministic test.

gh-xj (changes requested):
The intent here looks right: an explicit `limit` should win over the env/config default. I’m still blocking on two issues before merge.

First, this patch drops the existing fallback 128-turn ceiling for `memory.window` when `limit` is omitted. In `crates/app/src/memory/sqlite.rs`, only the explicit `payload.limit` path is clamped; the `default_window_size()` fallback is now unbounded. Because `config.memory.sliding_window` is exported directly into `LOONGCLAW_SLIDING_WINDOW` and `memory.sliding_window` has no upper-bound validation, fallback reads can now exceed 128. I think the fix is to resolve the source first, then clamp once.

Second, the new regression tests mutate `LOONGCLAW_SLIDING_WINDOW` globally without restoring prior state or serializing access. That makes them flaky under parallel test execution and conflicts with the deterministic env-handling pattern already used elsewhere in the repo. Please snapshot/restore the env var under a guard, or consolidate the scenarios into one deterministic test.

chumyin:
hi @kurozee @gh-xj, i took a fresh pass through this against current dev. the branch is still conflicting with dev, maintainer edits are disabled, and the memory window path has moved on quite a bit since this was opened. current dev already validates memory.sliding_window against the adapter cap, routes explicit extended window requests via allow_extended_limit, and has regression coverage for explicit-limit precedence, fallback behavior, and the 128 cap in crates/app/src/memory/tests.rs. i also re-ran that current regression coverage locally, and it passes on dev. because of that, i am closing this as superseded by the later mainline implementation instead of replaying an older diff onto the new memory architecture. thanks again for the original fix direction here.

PR #40, opened 2026-03-12
feat: close the external skills managed runtime loop

### LINKED ISSUE #39: Close the external skills managed runtime loop on alpha-test
## Problem

`alpha-test` already had `external_skills.policy` and `external_skills.fetch`, but the runtime still stopped short of a usable install/invoke loop. There was no managed local installation model, no inventory/index for installed skills, no direct invoke path, and no prompt-side exposure of installed skills.

That left external skills in a half-wired state: downloads could be guarded, but the runtime could not safely complete the full `fetch -> install -> list -> inspect -> invoke -> remove` lifecycle.

## Scope

- add a managed install root and installed skill index
- add lifecycle tools for install/list/inspect/invoke/remove
- expose installed skills in capability snapshots when enabled
- wire install root and auto-expose settings through config and runtime env
- update migration warnings and docs to describe the explicit runtime lifecycle
- cover the runtime gate and provider/system-prompt exposure with regression tests

## Notes

This is a follow-up to the earlier policy hardening work in #26, but it covers a different gap: closing the local runtime loop after download.

chumyin:
## Summary
- add a managed external-skills lifecycle with install, list, inspect, invoke, and remove tools
- wire install-root and auto-expose settings through config, runtime config, capability snapshots, and provider system prompts
- tighten runtime gating so disabled external skills no longer install, invoke, or auto-expose, and document the explicit lifecycle in migration warnings and READMEs

## Validation
- [x] cargo fmt --check
- [x] cargo test

Closes #39

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "adler2"
+version = "2.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+
+[[package]]
 name = "aes"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -665,10 +671,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "28dea519a9695b9977216879a3ebfddf92f1c08c05d984f8996aecd6ecdc811d"
 
 [[package]]
+name = "filetime"
+version = "0.2.27"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f98844151eee8917efc50bd9e8318cb963ae8b297431495d3f758616ea5c57db"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "libredox",
+]
+
+[[package]]
 name = "find-msvc-tools"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
+
+[[package]]
+name = "flate2"
+version = "1.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "843fba2746e448b37e26a819579957415c8cef339bf08564fe8b7ddbd959573c"
+dependencies = [
+ "crc32fast",
+ "miniz_oxide",
+]
 
 [[package]]
 name = "fnv"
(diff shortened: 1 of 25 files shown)


gh-xj (changes requested):
I re-ran this review adversarially to try to knock down the current findings. The core issues still stand; I'd block on these:

1. Modular builds are broken when `tool-file` is not enabled.
`external_skills.install` unconditionally calls the file-path resolver used by the file tool, but that helper is behind `#[cfg(feature = "tool-file")]`. Repro:
`cargo check -p loongclaw-app --no-default-features --features "provider-openai"`
This currently fails with `E0425`.

2. `external_skills.invoke` still does not close the managed-runtime loop.
The invoked `SKILL.md` instructions are returned from the tool, but they are only fed back into the model as ordinary tool-result / assistant text. They do not become privileged orchestration context, so the runtime behavior still depends on the model treating prior tool output as instructionally authoritative.

3. Large invoked skill payloads are still not reliably preserved across real follow-up paths.
The PR increases the `TurnEngine` envelope allowance for `external_skills.invoke`, but the multi-round `turn_loop` path still re-budgets tool results under the standard 8k/20k follow-up caps. So the “keep invoke payload intact” fix is only partial.

4. `replace=true` is non-atomic and can destroy a working install.
The existing managed install is removed before the replacement copy, validation, and index persistence complete. Any failure in that window leaves the previous working skill gone.

Non-blocking but worth fixing in the same series:

5. `setup` claims success without bootstrapping SQLite memory in the default daemon build.
The bootstrap block is behind the daemon's `memory-sqlite` feature, but that feature is not enabled by the default daemon feature set, so the success path is misleading.

6. `--apply-external-skills-plan` persists absolute local filesystem paths into durable prompt memory.
That leaks workstation layout / usernames to the provider via `profile_note`.

7. The documented `--setup` quickstart path is a rerun footgun.
This is narrower than a general install regression: the overwrite only happens when the user explicitly reuses the optional `--setup` fast path, which calls `setup --force`.

Given the above, I'm keeping this at `Needs Changes`.

chumyin:
Rebased onto the latest `alpha-test` and addressed the review blockers end to end.

What changed:
- Fixed the modular build break by making the shared file-path safety helpers available to `external_skills.install` even when `tool-file` is not enabled.
- Promoted `external_skills.invoke` follow-up instructions into trusted system context in both the fast-lane loop and the safe-lane coordinator, so invoke payloads are no longer re-budgeted or carried forward as ordinary assistant tool-result text.
- Made managed replace installs rollback-safe: stage incoming content, move the previous install aside, promote atomically, and restore the previous install/index if activation or index persistence fails.
- Stopped durable profile-note addenda from embedding absolute local filesystem paths; they now record artifact labels instead.
- Kept SQLite bootstrap available in the default daemon MVP path and verified it through the current onboarding flow on the rebased branch.
- Updated installer quickstart behavior to use `onboard` without `--force`, so rerunning the install helper stops on an existing config instead of overwriting it.
- After rebasing, fixed the additional no-default-features governor path references and clippy blockers that the newer base exposed.

Regression coverage / validation on the current PR head:
- `cargo fmt --check`
- `cargo clippy -p loongclaw-app --all-targets -- -D warnings`
- `cargo test`
- `cargo check -p loongclaw-app --no-default-features --features "provider-openai"`
- `cargo run -q -p loongclaw-daemon --bin loongclaw -- onboard --output <tmp>/loongclaw.toml --non-interactive --accept-risk --provider openai --api-key test-key --skip-model-probe`
- Re-ran the same `onboard` command against the same output path without `--force` and confirmed it fails with `config ... already exists (use --force to overwrite)`.

I will merge once the checks for `06b0da6` are green.

chumyin:
Checks are green on `06b0da6`. Merged into `alpha-test` with admin rebase under explicit operator authorization after the rebase follow-up fixes and validation above.

PR #79, opened 2026-03-13
feat(kernel): add additive builder/runtime migration surface

gh-xj:
## Summary

- add additive `kernel::KernelBuilder<P>` and `kernel::Kernel<P>` surfaces without removing `LoongClawKernel<P>`
- add `spec::BootstrapBuilder` as a new bootstrap entrypoint while preserving the legacy `spec::KernelBuilder`
- let `Kernel<P>` dereference to the legacy `&LoongClawKernel<_>` helper surface so existing read-only helpers keep working during migration
- add send/sync, concurrency, and bootstrap compatibility regression coverage

## Scope

- [x] Small and focused
- [ ] Includes docs updates (PR contract + migration guidance)
- [x] No unrelated refactors

## Risk Track

Touches the central kernel/bootstrap API used across the workspace, but keeps the exported surface backward compatible. Key invariants:
- `LoongClawKernel<P>` remains available for existing callers.
- `Kernel<P>` is an additive frozen runtime handle.
- `Kernel<P>` still bridges into legacy read-only helper signatures through `Deref` / `AsRef`.
- `loongclaw-spec::KernelBuilder` stays intact; `BootstrapBuilder` is additive.

## Breaking Change / Migration

This PR is **not** a breaking public API change.

Migration notes:
- existing `LoongClawKernel<P>` and `loongclaw-spec::KernelBuilder` call sites continue to work unchanged
- callers can opt into `KernelBuilder<P>::build()` to obtain the new `Kernel<P>` runtime handle
- callers can adopt `BootstrapBuilder::build()` for the new spec bootstrap path
- callers that still need extra registration before freeze can use `BootstrapBuilder::into_builder()` and then call `.build()`

## Validation

- [x] targeted regression: `cargo test -p loongclaw-kernel kernel_is_usable_from_concurrent_tasks`
- [x] targeted regression: `cargo test -p loongclaw-spec kernel_bootstrap::tests::`
- [x] branch synced with current `alpha-test`

## Linked Issues

Architecture audit C5

### DIFF
--- a/crates/kernel/src/kernel.rs
+++ b/crates/kernel/src/kernel.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::{BTreeMap, BTreeSet},
+    ops::Deref,
     sync::{
         Arc,
         atomic::{AtomicU64, Ordering},
@@ -69,6 +70,17 @@ pub struct LoongClawKernel<P: PolicyEngine> {
     clock: Arc<dyn Clock>,
     audit: Arc<dyn AuditSink>,
     event_seq: AtomicU64,
+}
+
+/// Additive migration alias for the legacy kernel surface.
+///
+/// During the compatibility window this still exposes the executable
+/// `LoongClawKernel` API, while also supporting `.build()` into the new
+/// frozen `Kernel<P>` handle.
+pub type KernelBuilder<P> = LoongClawKernel<P>;
+
+pub struct Kernel<P: PolicyEngine> {
+    inner: LoongClawKernel<P>,
 }
 
 impl<P: PolicyEngine> LoongClawKernel<P> {
@@ -193,6 +205,11 @@ impl<P: PolicyEngine> LoongClawKernel<P> {
         self.memory_plane
             .set_default_core_adapter(name)
             .map_err(KernelError::from)
+    }
+
+    #[must_use]
+    pub fn build(self) -> Kernel<P> {
+        Kernel { inner: self }
     }
 
     pub fn issue_token(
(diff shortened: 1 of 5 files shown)


chumyin (changes requested):
Requesting changes because this refactor currently lands as a public API break while the PR still presents it as an incremental migration path.

- `LoongClawKernel<P>` no longer preserves the old executable-kernel surface. It now aliases `KernelBuilder<P>`, while runtime methods live on `Kernel<P>`. That means downstream code still typing values, parameters, or fields as `LoongClawKernel<_>` and calling `issue_token` / `execute_*` now fails to compile.
- `loongclaw-spec` also changes its public construction flow in a breaking way: the exported `KernelBuilder` name is replaced by `BootstrapBuilder`, and `build()` now freezes immediately into `Kernel`, so previous `build(); register_*(); set_default_*()` usage no longer type-checks.

If this is intended to be a breaking change, please update the PR description and migration guidance to say that explicitly. If the goal is truly incremental migration, this needs a real compatibility layer rather than aliasing the old public name to the builder type.

I did not find another concrete runtime regression beyond these API contract issues, and the in-repo tests/CI look good.

gh-xj:
Addressed the requested changes on top of a rebase onto `alpha-test`:

- removed the public `LoongClawKernel<P>` alias and migrated in-repo call sites to explicit `KernelBuilder<P>` / `Kernel<P>` usage
- updated the PR description to treat this as an intentional pre-1.0 breaking change, with a dedicated migration section for the kernel/bootstrap surface
- routed `SecurityScanEvaluated` through `Kernel::record_audit_event(...)` so security-scan summaries use the same kernel event sequencing as the rest of the audit stream
- refreshed the architecture drift report after the rebase

Local verification on the rebased branch:
- `cargo fmt --all --check`
- `cargo test --workspace`

CI is now rerunning on commit `49d6b6e`.

gh-xj:
Follow-up on `bece243`:

- preserve `SecurityScanEvaluated` on builder-phase blocked exits that happen after scan evaluation but before the main success-path emit
- cover the late blocked path with a new regression: non-empty security scan report + invalid default core selection still retains the audit summary in `audit_events`
- refresh the generated architecture drift report after the additional `spec_execution.rs` changes

Local verification on this follow-up commit:
- `cargo fmt --all --check`
- `cargo test --workspace`

gh-xj:
@chumyin Addressed the change-request items on the current head `bece243`:

- the PR body now states this is an intentional pre-1.0 breaking change and includes migration guidance for the `KernelBuilder` / `Kernel` split and `BootstrapBuilder` flow
- the public `LoongClawKernel<P>` alias has been removed from the exported surface
- the follow-up commit preserves `SecurityScanEvaluated` on late builder-phase blocked exits and adds regression coverage for that path
- local verification and the full GitHub check matrix are green on the updated branch

Requesting another review pass on the current head.

chumyin (approved):
hi @gh-xj, thanks for sticking with this one. i rebased the branch onto current `alpha-test` and tightened the change into an additive migration surface so we keep `LoongClawKernel<P>` / `loongclaw-spec::KernelBuilder` compatibility while still landing `KernelBuilder<P>`, `Kernel<P>`, and `BootstrapBuilder`. local validation plus the full github check matrix are green now. congrats, i’m merging this.

PR #84, opened 2026-03-13
refactor(channel): let registry descriptors drive snapshots

chumyin:
## Summary
- let each channel registry descriptor own its snapshot builder instead of relying on a central platform match
- reuse descriptor metadata for catalog rendering and platform normalization
- accept the displayed `Feishu/Lark` label when normalizing channel platform input

## Why
- keeps loongclaw closer to the metadata-driven channel registry shape used by openclaw while staying lightweight on alpha-test
- reduces the number of places that must change when adding future channels or high-quality stubs
- avoids pulling the larger integration-branch abstractions into this small follow-up PR

## Validation
- `<local-absolute-path> test -p loongclaw-app channel::registry::tests:: --all-features --target-dir target/codex-alpha-registry-descriptors-pr`
- `<local-absolute-path> clippy -p loongclaw-app --all-targets --all-features --target-dir target/codex-alpha-registry-descriptors-pr -- -D warnings`
- `<local-absolute-path> fmt --all --check`
- `<local-absolute-path> test --workspace --all-features --target-dir target/codex-alpha-registry-descriptors-pr -- --test-threads=1`

PR #85, opened 2026-03-13
refactor(channel): drive doctor checks from registry metadata

chumyin:
## Summary
- move channel doctor check labels out of `crates/daemon/src/doctor_cli.rs` and into app-level channel registry metadata
- expose a registry lookup helper so daemon doctor checks follow channel operation metadata instead of a local `(channel_id, operation_id)` match table
- make Feishu send doctor output more explicit as `feishu direct send`, keeping runtime names on the operation metadata path for future channels and stubs

## Why
- adding a new channel surface should not require updating both registry metadata and daemon hardcoded doctor labels
- this keeps channel capability metadata closer to the provider registry boundary, which is a better fit for future high-quality stub channels
- it is a small follow-up to PR #84 without pulling in heavier abstractions

## Validation
- `<local-absolute-path> test -p loongclaw-daemon build_channel_surface_checks_uses_operation_specific_doctor_name_for_feishu_send --all-features --target-dir <local-absolute-path>`
- `<local-absolute-path> test -p loongclaw-app channel::registry::tests:: --all-features --target-dir <local-absolute-path>`
- `<local-absolute-path> test -p loongclaw-daemon build_channel_surface_checks_ --all-features --target-dir <local-absolute-path>`
- `<local-absolute-path> fmt --all --check`
- `<local-absolute-path> clippy -p loongclaw-app -p loongclaw-daemon --all-targets --all-features --target-dir <local-absolute-path> -- -D warnings`
- `<local-absolute-path> test --workspace --all-features --target-dir <local-absolute-path> -- --test-threads=1`
- `./scripts/check_architecture_drift_freshness.sh docs/releases/architecture-drift-2026-03.md`

### DIFF
--- a/crates/app/src/channel/mod.rs
+++ b/crates/app/src/channel/mod.rs
@@ -40,9 +40,9 @@ mod runtime_state;
 mod telegram;
 
 pub use registry::{
-    ChannelCatalogEntry, ChannelCatalogOperation, ChannelOperationHealth, ChannelOperationStatus,
-    ChannelStatusSnapshot, channel_status_snapshots, list_channel_catalog,
-    normalize_channel_platform,
+    ChannelCatalogEntry, ChannelCatalogOperation, ChannelDoctorCheckSpec, ChannelOperationHealth,
+    ChannelOperationStatus, ChannelStatusSnapshot, channel_doctor_check_spec,
+    channel_status_snapshots, list_channel_catalog, normalize_channel_platform,
 };
 pub use runtime_state::ChannelOperationRuntime;
 use runtime_state::ChannelOperationRuntimeTracker;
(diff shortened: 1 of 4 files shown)


chumyin:
Superseded by the current channel registry-first stack rooted at #86 and extended through #109.

Why this is being closed:
- the main stack is already anchored on `alpha-test` at #86
- the later stack carries the doctor-metadata direction forward and then keeps evolving it through grouped surfaces, registry-level doctor specs, operation descriptors, and requirement metadata
- keeping this earlier overlapping PR open would leave two competing doctor/channel-refactor entry points against the same base branch

Replacement chain:
#86 -> #89 -> #90 -> #93 -> #96 -> #97 -> #98 -> #99 -> #100 -> #101 -> #102 -> #103 -> #104 -> #105 -> #106 -> #108 -> #109

PR #113, opened 2026-03-14
fix(provider): parse inline function tool calls

gh-xj:
## Summary

- Parse inline `<function=...>` blocks from assistant text into `ToolIntent`s when the provider does not return structured `message.tool_calls`.
- Recover JSON-typed inline parameter values when the provider emits quoted strings, booleans, numbers, arrays, or objects inside `<parameter=...>` blocks.
- Record structured provider parse telemetry in `raw_meta` for recovered and malformed inline function blocks so visible tool-call fallbacks are inspectable instead of silent.

## Scope

- [x] Small and focused
- [x] No unrelated refactors

## Risk Track

- [x] Track A (routine/low-risk)
- [ ] Track B (higher-risk/policy-impacting)

## Validation

Additional checks:
- `cargo test -p loongclaw-app extract_provider_turn_ -- --nocapture`
- `cargo test -p loongclaw-app provider_tool_aliases_flow_through_parse_and_turn_validation -- --nocapture`
- `task verify`

## Linked Issues

Closes #

### DIFF
--- a/crates/app/src/provider/shape.rs
+++ b/crates/app/src/provider/shape.rs
@@ -903,11 +903,64 @@ mod tests {
     }
 
     #[test]
+    fn extract_provider_turn_does_not_execute_multiline_indented_code_block_examples() {
+        let body = serde_json::json!({
+            "choices": [{
+                "message": {
+                    "content": "示例：\n\n    第一步\n    <function=shell.exec><parameter=command>ls</parameter></function>"
+                }
+            }]
+        });
+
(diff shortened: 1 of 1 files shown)


chumyin (approved):
Verified the latest inline-function fallback refinement on PR head 22a4d1d. The remaining Markdown indented-code-block false positive is addressed, and the added regression coverage now locks literal examples and valid indented tool calls separately.

PR #137, opened 2026-03-14
fix(approval): preserve approval-required tool failures

### LINKED ISSUE #135: [BUG] Alpha-test rewrites shell approval-required failures as misleading assistant narration
## Summary

On `alpha-test`, a shell command can be explicitly allowlisted in app config and still be blocked by the kernel's default approval policy. The conversation layer currently loses that distinction and can rewrite approval-required failures as generic assistant narration, which makes the system sound broken instead of truthfully reporting that approval is still required.

## Expected behavior

- `PolicyError::ToolCallApprovalRequired` remains distinct from a generic policy denial through the conversation stack.
- approval-required shell failures surface as approval-required, not as a misleading assistant paraphrase.
- regression coverage proves that allowlisting `curl` does not bypass the kernel's default shell approval requirement.

## Reproduction steps

1. Start from `alpha-test`.
2. Configure `tools.shell_allowlist = ["curl", "jq", "ls", "grep", "cat"]`.
3. Ask the assistant to use `shell.exec` with `curl`.
4. Observe that the assistant identifies approval is required.
5. Provide explicit approval in chat.
6. Observe that execution still does not proceed and the assistant can report misleading narration instead of a truthful approval-required outcome.

## Version / commit

`alpha-test` branch head

## Contribution track impact

chumyin:
## Summary

- What changed? Preserved kernel `ToolCallApprovalRequired` outcomes as `ApprovalRequired` instead of collapsing them into generic policy denial; `shell.exec` approval-required failures now return `TurnResult::NeedsApproval(...)` and bypass the completion-pass rewrite path; safe-lane plan-node mapping preserves the same semantics and regression tests cover allowlisted `curl` still requiring approval.
- Why this change is needed? `alpha-test` was rewriting approval-required shell failures into misleading assistant narration, especially in allowlisted `curl` local-search workflows, even though `tools.shell_allowlist` and kernel approval policy are separate gates.

## Scope

- [x] Small and focused
- [x] No unrelated refactors

No docs update was needed for this internal behavior fix.

## Risk Track

- This changes policy-impacting failure classification and reply surfacing, but does not relax any shell execution guardrails.
- Actual kernel enforcement is unchanged; only approval-required outcomes now keep their semantics through fast-lane and safe-lane execution paths.
- Adjacent policy-denied and tool-error paths retain existing completion-pass behavior and are covered by regression checks.

## Validation

Additional scenario checks covered the allowlisted `curl` approval-required regression, direct approval-required reply finalization, and nearby denial/error control paths.

Not applicable for the config/env fallback item: this change does not alter config/env fallback behavior, limits, or defaults.

Not applicable for the process-global env item: the added tests do not mutate process-global environment state.

## Linked Issues

Closes #135

### DIFF
--- a/crates/app/src/conversation/plan_executor.rs
+++ b/crates/app/src/conversation/plan_executor.rs
@@ -35,6 +35,7 @@ pub enum PlanRunFailure {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PlanNodeErrorKind {
     Retryable,
+    ApprovalRequired,
     PolicyDenied,
     NonRetryable,
 }
@@ -56,6 +57,13 @@ impl PlanNodeError {
     pub fn policy_denied(message: impl Into<String>) -> Self {
         Self {
             kind: PlanNodeErrorKind::PolicyDenied,
+            message: message.into(),
+        }
+    }
+
+    pub fn approval_required(message: impl Into<String>) -> Self {
+        Self {
+            kind: PlanNodeErrorKind::ApprovalRequired,
             message: message.into(),
         }
     }
--- a/crates/app/src/conversation/plan_executor.rs
+++ b/crates/app/src/conversation/plan_executor.rs
@@ -244,7 +244,11 @@ impl PlanExecutor {
                                 error_kind: Some(normalized.kind),
                                 error: Some(normalized.message.clone()),
                             });
-                            if attempt == node.max_attempts {
+                            // Approval-required failures are terminal until the caller changes
+                            // execution context, so retrying the same node cannot make progress.
+                            if normalized.kind == PlanNodeErrorKind::ApprovalRequired
+                                || attempt == node.max_attempts
+                            {
                                 let elapsed_ms = started_at.elapsed().as_millis();
                                 return PlanRunReport {
                                     status: PlanRunStatus::Failed(PlanRunFailure::NodeFailed {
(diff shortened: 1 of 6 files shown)


crates/app/src/conversation/plan_executor.rs
```diff
@@ -35,6 +35,7 @@ pub enum PlanRunFailure {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PlanNodeErrorKind {
     Retryable,
+    ApprovalRequired,
```
chumyin, line 38: Introducing `ApprovalRequired` here is not sufficient by itself, because `PlanExecutor::execute()` still retries every `Err(PlanNodeError)` until `attempt == node.max_attempts`. In the safe-lane path, tool nodes inherit `safe_lane_node_max_attempts`, whose default is still `2`, so an approval-required tool call will be executed twice before the turn finally returns `NeedsApproval`. That means extra latency, an extra consumed attempt budget, and a duplicate denial audit record for a condition that should terminate immediately on the first hit. This new error kind needs to short-circuit the node retry loop the same way a terminal failure would.

crates/app/src/conversation/tests.rs
```diff
@@ -4888,6 +4888,66 @@ async fn handle_turn_with_runtime_tool_failure_completion_error_uses_raw_reason_
...
+            }],
+            raw_meta: Value::Null,
+        }),
+        Ok("MODEL_SHOULD_NOT_RUN".to_owned()),
+    );
+
+    let coordinator = ConversationTurnCoordinator::new();
+    let reply = coordinator
+        .handle_turn_with_runtime(
+            &test_config(),
```
chumyin, line 4919: This regression test is useful, but it still uses `test_config()`, and `ConversationConfig::default()` keeps `safe_lane_plan_execution_enabled = false`. So the test never exercises the safe-lane plan-executor path changed in this PR; it only proves that the non-plan runtime flow preserves the approval-required reply and skips the completion rewrite. Given the bug being fixed here, I think we still need an end-to-end safe-lane test with plan execution enabled, ideally also covering the default `safe_lane_node_max_attempts = 2` case so repeated approval-required attempts cannot regress silently.

chumyin (commented):
I found one functional issue in the safe-lane approval flow and one related test coverage gap.

crates/app/src/conversation/tests.rs
chumyin: Covered in 43f7c1b. I added `conversation::tests::handle_turn_with_runtime_safe_lane_approval_required_is_not_retried`, which enables `safe_lane_plan_execution_enabled`, keeps the default node retry budget, and asserts that the approval-required path finalizes without a completion rewrite while recording exactly one authorization-denied audit event.

crates/app/src/conversation/plan_executor.rs
chumyin: Fixed in 43f7c1b. `PlanExecutor::execute()` now treats `PlanNodeErrorKind::ApprovalRequired` as terminal and returns immediately instead of consuming the remaining node retries. I also added `conversation::plan_executor::tests::executor_stops_retrying_after_approval_required_failure` to lock the short-circuit behavior.

gh-xj (approved):

PR #149, opened 2026-03-15
feat: adopt discovery-first provider tool routing

### LINKED ISSUE #148: [FEATURE] Adopt a discovery-first provider tool architecture
## Problem statement

`alpha-test` still exposes a broad static provider tool surface in two places at once: the JSON function schema and the system-prompt capability snapshot. That creates three concrete problems.

1. The model pays a token tax on every turn for tools it will not use.
2. Non-core tools such as `file.read`, `file.write`, `shell.exec`, and the managed external-skills lifecycle can still be guessed or directly called instead of being discovered intentionally.
3. The current provider/runtime path has no app-native discovery-first contract even though the product direction is to let the model start with only a minimal core tool surface and use search to reach everything else.

There is also architectural drift around governance and disclosure:

- `crates/app/src/tools/mod.rs::provider_tool_definitions()` still hardcodes the full provider-visible schema.
- `crates/app/src/provider/request_message_runtime.rs` still appends a full capability snapshot to the system prompt.
- `crates/app/src/conversation/turn_engine.rs` and the coordinator safe-lane path still validate direct provider tool calls against the broad static known-tool set.
- installed external skills still default to auto-exposure in prompt-level disclosure.
- `crates/spec` has a `tool_search`, but it is not on the current app/provider hot path.

The result is that LoongClaw has the pieces for a stronger tool architecture, but not the discovery-first runtime contract itself.

## Proposed solution

Adopt a discovery-first hybrid tool architecture for the app/provider runtime.

The first implementation slice should:

- expose only `tool_search` and `tool_invoke` as provider-callable function tools
- introduce an app-native runtime catalog that separates provider-core tools from discoverable tools
- make `tool_search` return compact tool cards plus a short-lived invoke lease instead of full per-tool schemas
- make `tool_invoke` validate that lease and dispatch the underlying non-core tool through the existing kernel-governed tool plane
- reject direct provider calls to discoverable tools such as `file.read` or `shell.exec`
- shrink the system-prompt capability snapshot so it describes only the discovery contract instead of the full tool inventory
- disable installed external-skill auto-exposure by default
- close the permissive `execute_tool(..., None)` helper path so kernel-first is true in code as well as in docs

This keeps the provider-visible surface minimal, reduces tokens, preserves one governed execution path, and leaves room for later progressive exposure modes such as dispatcher-only, ephemeral-stub, and ephemeral-full.

Design notes for this slice are already captured in:

- `docs/plans/2026-03-15-tool-discovery-architecture-design.md`
- `docs/plans/2026-03-15-tool-discovery-architecture.md`

## Alternatives considered

1. Keep the current static full-surface schema.
This is the smallest immediate change, but it keeps the token overhead, weakens the search-first runtime story, and allows direct guessing of non-core tools.

2. Search first and then dynamically inject full per-tool schemas into the next provider turn.
This improves direct function-calling ergonomics, but it increases control-plane complexity and creates more parser/schema drift risk before the core discovery contract is stable.

3. Add a database-backed or vector-DB-backed tool search layer now.
This may help ranking later, but it is unnecessary for the first shipped slice. A derived in-memory catalog plus signed short-lived leases is lighter and sufficient for the current tool surface.

## Estimated risk track

This change affects provider disclosure, tool routing, and runtime governance boundaries, so it should land behind a well-documented design note with regression coverage for fast-lane, safe-lane, prompt disclosure, and lease validation.

chumyin:
## Summary

- What changed?
  - limit the provider-visible tool schema to `tool_search` and `tool_invoke`, backed by a runtime tool catalog and short-lived discovery leases for non-core tools
  - route wrapped tool invocations through effective capability derivation and effective shell-policy enforcement, and bind leases to token/session/turn scope when runtime context is available
  - default installed external skills to no auto-exposure, shrink the prompt capability snapshot to the discovery contract, and add design docs plus regression coverage for the new routing path
  - add provider-shape end-to-end follow-up coverage across OpenAI chat-completions, Responses, Anthropic native content blocks, Bedrock Converse blocks, and inline-function parsing for the real `tool.search -> tool.invoke` coordinator path
  - emit and summarize discovery-first runtime telemetry for search rounds, follow-up provider turns, raw-output preservation, token delta estimates, and search-to-invoke hit rate, and expose that summary through session-history helpers
  - add a durable discovery-first runtime contract note under `docs/design-docs/` so provider-core tool exposure, lease expectations, compatibility rewrites, and follow-up turn requirements stop drifting
- Why this change is needed?
  - reduce per-turn schema/prompt token overhead
  - force discovery-first access to non-core tools without opening a second ungoverned execution path
  - harden the new routing path with real provider-shape coverage and lightweight observability before merge

## Scope

- [ ] Small and focused
- [x] No unrelated refactors

## Risk Track

- This changes provider disclosure, tool routing, and default governance behavior. The main guardrails are lease validation, direct-call rejection for discoverable tools, wrapped capability/shell-policy enforcement, and regression coverage across fast-lane and safe-lane execution.
- Before: the provider schema and prompt-level capability snapshot exposed the broad static tool surface, installed external skills auto-exposed by default, and wrapped `tool.invoke(shell.exec)` payloads could miss effective shell-policy evaluation.
- After: the provider sees only the discovery contract, installed external skills stay hidden unless explicitly surfaced, wrapped shell calls are evaluated under the same allow/deny/default policy as direct shell execution, and the real follow-up path is exercised across provider response families with telemetry on extra provider-round cost.

## Validation

- [ ] `cargo test --workspace --all-features`
- [ ] Additional scenario/benchmark checks (if applicable)

Additional verification:

- `cargo test -p loongclaw-app summarize_discovery_first_events -- --nocapture`
- `cargo test -p loongclaw-app provider_shape_tool_search -- --nocapture`
- `cargo test -p loongclaw-app handle_turn_with_runtime_tool_search_ -- --nocapture`
- `cargo test -p loongclaw-app -- --test-threads=1`
- regression coverage for wrapped `tool.invoke` capability derivation, wrapped shell policy enforcement, lease replay rejection across turn/session scope, discovery-first provider-shape follow-up execution, raw-output preservation, and checkpoint-visible follow-up context
- process-global env mutation remains serialized and restored via `ScopedEnv` in [`crates/app/src/test_support.rs`](./crates/app/src/test_support.rs)

## Linked Issues

Closes #148

## Summary by CodeRabbit

* **New Features**
  * Implemented discovery-first tool runtime with searchable tool catalog and lease-based invocation.
  * Added tool categorization distinguishing provider-core tools from discoverable tools.
  * Enhanced follow-up handling for multi-round provider interactions.
  * Added discovery-first event tracking and analytics summaries.

* **Chores**
  * Changed default for external skills auto-exposure from enabled to disabled.
  * Updated tool governance and visibility rules.

### DIFF
--- a/crates/app/src/channel/feishu/webhook.rs
+++ b/crates/app/src/channel/feishu/webhook.rs
@@ -1099,6 +1099,22 @@ mod tests {
                             Json(json!({
                                 "choices": [{
                                     "message": {
+                                        "content": "looking up the right callback card tool",
+                                        "tool_calls": [{
+                                            "id": "call_tool_search_1",
+                                            "type": "function",
+                                            "function": {
+                                                "name": "tool_search",
+                                                "arguments": "{\"query\":\"feishu card update callback token markdown\",\"limit\":1}"
+                                            }
+                                        }]
+                                    }
+                                }]
+                            }))
+                        } else if *turn_index == 2 {
+                            Json(json!({
+                                "choices": [{
+                                    "message": {
                                         "content": "updating card",
                                         "tool_calls": [{
                                             "id": "call_feishu_card_update_1",
--- a/crates/app/src/config/conversation.rs
+++ b/crates/app/src/config/conversation.rs
@@ -200,6 +200,8 @@ pub struct ConversationTurnLoopConfig {
     pub max_followup_tool_payload_chars: usize,
     #[serde(default = "default_turn_loop_max_followup_tool_payload_chars_total")]
     pub max_followup_tool_payload_chars_total: usize,
+    #[serde(default = "default_turn_loop_max_discovery_followup_rounds")]
+    pub max_discovery_followup_rounds: usize,
 }
 
 impl Default for ConversationTurnLoopConfig {
@@ -213,6 +215,7 @@ impl Default for ConversationTurnLoopConfig {
             max_followup_tool_payload_chars: default_turn_loop_max_followup_tool_payload_chars(),
             max_followup_tool_payload_chars_total:
                 default_turn_loop_max_followup_tool_payload_chars_total(),
+            max_discovery_followup_rounds: default_turn_loop_max_discovery_followup_rounds(),
         }
     }
 }
@@ -436,6 +439,10 @@ const fn default_turn_loop_max_followup_tool_payload_chars() -> usize {
 
 const fn default_turn_loop_max_followup_tool_payload_chars_total() -> usize {
     20_000
+}
+
+const fn default_turn_loop_max_discovery_followup_rounds() -> usize {
+    2
 }
 
 const fn default_fast_lane_max_tool_steps_per_turn() -> usize {
--- a/crates/app/src/config/runtime.rs
+++ b/crates/app/src/config/runtime.rs
@@ -2187,7 +2187,7 @@ api_key = "${DEEPSEEK_API_KEY}"
         assert!(raw.contains("[external_skills]"));
         assert!(raw.contains("enabled = false"));
         assert!(raw.contains("require_download_approval = true"));
-        assert!(raw.contains("auto_expose_installed = true"));
+        assert!(raw.contains("auto_expose_installed = false"));
 
         let (_, loaded) = load(Some(&path_string)).expect("config load should pass");
         assert!(!loaded.external_skills.enabled);
@@ -2195,7 +2195,7 @@ api_key = "${DEEPSEEK_API_KEY}"
         assert!(loaded.external_skills.allowed_domains.is_empty());
         assert!(loaded.external_skills.blocked_domains.is_empty());
         assert!(loaded.external_skills.install_root.is_none());
-        assert!(loaded.external_skills.auto_expose_installed);
+        assert!(!loaded.external_skills.auto_expose_installed);
 
         let _ = fs::remove_file(path);
     }
--- a/crates/app/src/config/tools_memory.rs
+++ b/crates/app/src/config/tools_memory.rs
@@ -511,7 +511,7 @@ const fn default_true() -> bool {
 }
 
 const fn default_auto_expose_installed() -> bool {
-    true
+    false
 }
 
 fn normalize_domain_entries(entries: &[String]) -> Vec<String> {
@@ -706,7 +706,7 @@ child_tool_allowlist = ["file.read", "shell.exec"]
         assert!(config.allowed_domains.is_empty());
         assert!(config.blocked_domains.is_empty());
         assert!(config.install_root.is_none());
-        assert!(config.auto_expose_installed);
+        assert!(!config.auto_expose_installed);
     }
 
     #[test]
--- a/crates/app/src/conversation/analytics.rs
+++ b/crates/app/src/conversation/analytics.rs
@@ -392,6 +392,25 @@ pub struct ConversationEventRecord {
     pub payload: Value,
 }
 
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DiscoveryFirstEventSummary {
+    pub search_round_events: u32,
+    pub followup_requested_events: u32,
+    pub followup_result_events: u32,
+    pub raw_output_followup_events: u32,
+    pub search_to_invoke_hits: u32,
+    pub aggregate_added_estimated_tokens: u64,
+    pub added_estimated_token_samples: u32,
+    pub average_added_estimated_tokens: Option<u32>,
+    pub latest_followup_outcome: Option<String>,
+    pub latest_followup_tool_name: Option<String>,
+    pub latest_followup_target_tool_id: Option<String>,
+    pub latest_initial_estimated_tokens: Option<u32>,
+    pub latest_followup_estimated_tokens: Option<u32>,
+    pub latest_added_estimated_tokens: Option<u32>,
+    pub outcome_counts: BTreeMap<String, u32>,
+}
+
 pub fn parse_conversation_event(content: &str) -> Option<ConversationEventRecord> {
     let parsed = serde_json::from_str::<Value>(content).ok()?;
     if parsed.get("type")?.as_str()? != "conversation_event" {
@@ -428,6 +447,22 @@ where
     I: IntoIterator<Item = &'a str>,
 {
     summarize_safe_lane_history(contents).summary
+}
+
+pub fn summarize_discovery_first_events<'a, I>(contents: I) -> DiscoveryFirstEventSummary
+where
+    I: IntoIterator<Item = &'a str>,
+{
+    let mut summary = DiscoveryFirstEventSummary::default();
+
+    for content in contents {
+        let Some(record) = parse_conversation_event(content) else {
+            continue;
+        };
+        fold_discovery_first_event_record(&record, &mut summary);
+    }
+
+    summary
 }
 
 fn fold_safe_lane_event_record(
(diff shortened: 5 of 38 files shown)


chumyin:
i took a close pass here. the direction looks good, but i'm not merging this yet because there are two blocker-level gaps in the current provider path:

- discovery-first is not wired end to end in the coordinator path yet. the main provider flow still does `request_turn -> execute tools -> request_completion`, so after a first-round `tool.search` result there is no follow-up provider tool turn to turn that discovery result into `tool.invoke`. i reproduced this locally with a two-round provider scenario and the coordinator only requested one provider turn before jumping straight to completion.
- the real parser/runtime bridge is still inconsistent with the new contract. `provider::shape` still emits legacy direct tool intents like `file.read` / `shell.exec`, but the fast-lane and safe-lane executors now hard-reject anything that is not provider-core `tool.search` / `tool.invoke`. on top of that, real parsed tool intents still carry empty `session_id` / `turn_id`, so the new lease validation does not actually get real turn/session scope on the production provider path.

for reference, i re-ran `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features`, and `cargo test -p loongclaw-app -- --test-threads=1` locally, and they pass on the current head. i think this one needs either a real compatibility rewrite + scope propagation bridge, or the coordinator/provider loop needs to be extended so `tool.search -> tool.invoke` is actually executable before we land the discovery-first contract.

chumyin:
Addressed in `2182e10`.

This closes the two blocker-level gaps called out here: https://github.com/loongclaw-ai/loongclaw/pull/149#issuecomment-4062446663

- The real coordinator provider path now keeps the same outer `session_id` / `turn_id` scope and will request a follow-up provider turn after a `tool.search` round, so `tool.search -> tool.invoke` is now executable end to end instead of jumping straight from `request_turn -> execute tools -> request_completion`.
- The real parser/runtime bridge now rewrites discoverable legacy provider tool calls (`file.read`, `shell.exec`, inline function calls, Anthropic `tool_use`, Bedrock `toolUse`, etc.) into `tool.invoke`, and parsed `ToolIntent`s now carry the real production `session_id` / `turn_id` instead of empty strings.
- The follow-up provider turn path is bounded by the existing `conversation.turn_loop.max_rounds` budget and fails open to the current raw reply if the extra provider turn itself errors, so the discovery-first path gets the extra turn without regressing the current fallback behavior.

Added regression coverage for:
- scoped parser rewrite from legacy discoverable tools to `tool.invoke`
- coordinator follow-up provider turn after a first-round `tool.search`
- legacy alias parsing now validating against the provider-core contract (`tool.invoke` + lease scope) instead of the old direct-tool path

Re-ran locally:
- `cargo test -p loongclaw-app -- --test-threads=1`

All green on the updated head.

chumyin:
Pushed an updated head for this branch: `9603caf`.

This update closes the two remaining discovery-first blockers called out in review:

- Raw-output mode no longer short-circuits after a first-round `tool.search` payload. If the lane still requires a provider follow-up, raw mode now preserves the extra provider turn instead of terminating before `tool.invoke` can happen.
- Turn-checkpoint finalization now stays anchored to the original visible preparation context instead of the hidden synthetic follow-up context, so checkpoint repair/runtime probe no longer sees a false preparation fingerprint mismatch after a successful discovery-first follow-up.

I also kept the provider/runtime bridge aligned end-to-end:

- runtime/provider `request_turn` now propagates `session_id` and `turn_id` into the real provider path
- provider shaping rewrites discoverable tool calls to scoped `tool.invoke` consistently, including the Responses function-call path
- coordinator follow-up config reload still works after provider-switch turns under the wrapped `tool.invoke` form
- daemon import-cli tests were updated to the new scoped provider-turn API

Regression coverage added/updated on this branch includes:

- raw-output `tool.search` still requesting the follow-up provider turn
- follow-up checkpoint snapshots staying repairable against visible context
- provider Responses function calls normalizing to scoped `tool.invoke`
- coordinator/session/delegate tests updated to use discovery-first provider intents instead of legacy direct non-core provider calls

Local verification completed:

- `cargo fmt --all --check`
- `cargo test -p loongclaw-app -- --test-threads=1`

GitHub Actions for `9603caf` are now running; I'll follow up if any CI check reports a real regression.

chumyin:
Rebased this branch onto the latest `alpha-test` and re-verified the discovery-first path on the current head `96e206b`.

The follow-up review blockers that were raised earlier are covered on the rebased branch, and the only additional fix needed after rebase was test drift alignment in `crates/app/src/conversation/tests.rs`.

Fresh local validation completed on the current head:
- `cargo fmt --all --check`
- `cargo test -p loongclaw-app -- --test-threads=1`

All passed locally, and the GitHub checks for `96e206b8877899ffafa04203da2b93510df3b9a1` are green.

chumyin:
Pushed `b8fccff` to extend the discovery-first track with follow-up hardening.

This update adds:
- provider-shape end-to-end coverage for the real `tool.search -> tool.invoke` follow-up path across OpenAI chat-completions, Responses, Anthropic native tool blocks, Bedrock Converse tool blocks, and inline-function parsing
- persisted discovery-first telemetry plus analytics/session-history summaries for search rounds, extra provider turns, raw-output preservation, token delta estimates, and search-to-invoke hits
- a durable architecture note at `docs/design-docs/discovery-first-tool-runtime-contract.md`

Local verification on the new head:
- `cargo fmt --all --check`
- `cargo test -p loongclaw-app summarize_discovery_first_events -- --nocapture`
- `cargo test -p loongclaw-app provider_shape_tool_search -- --nocapture`
- `cargo test -p loongclaw-app handle_turn_with_runtime_tool_search_ -- --nocapture`
- `cargo test -p loongclaw-app -- --test-threads=1`

I also rechecked the staged diff locally for contract drift and did not find a blocker beyond the still-pending full CI run on the updated PR head.

chumyin:
Pushed  to extend the discovery-first track with follow-up hardening.

This update adds:
- provider-shape end-to-end coverage for the real  follow-up path across OpenAI chat-completions, Responses, Anthropic native tool blocks, Bedrock Converse tool blocks, and inline-function parsing
- persisted discovery-first telemetry plus analytics/session-history summaries for search rounds, extra provider turns, raw-output preservation, token delta estimates, and search-to-invoke hits
- a durable architecture note at 

Local verification on the new head:
- 
- 
running 2 tests
test conversation::analytics::tests::summarize_discovery_first_events_ignores_lookalikes_and_tracks_latest_request_snapshot ... ok
test conversation::analytics::tests::summarize_discovery_first_events_counts_followup_and_tokens ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1195 filtered out; finished in 0.02s
- 
- 
- 

I also rechecked the staged diff locally for contract drift and did not find a blocker beyond the still-pending full CI run on the updated PR head.

chumyin:
Pushed `beabf75` to address the remaining `rust-quality` blocker on this branch.

Root cause:
- CI runs clippy with `-D clippy::expect_used`, and the discovery-first app-tool bridge in `crates/app/src/conversation/turn_engine.rs` still had an `expect(...)` on the post-`tool.invoke` descriptor lookup.

Fix:
- Replaced that `expect(...)` with an explicit fail-closed `tool_catalog_drift` error path, so a catalog mismatch now returns a governed non-retryable tool failure instead of panicking.
- This keeps the discovery-first runtime semantics unchanged for valid requests while removing the remaining panic-shaped edge in the app-tool preflight path.

Fresh local verification on `beabf75`:
- `cargo fmt --all --check`
- `cargo test -p loongclaw-app -- --test-threads=1`

All passed locally. GitHub Actions for `beabf75` are now running.

gh-xj (approved):

PR #165, opened 2026-03-15
fix: harden alpha-test onboarding edge cases

### LINKED ISSUE #161: [Bug] Harden alpha-test onboarding edge cases
## Summary

Revalidated against `origin/alpha-test@66c1a99044f008acd0fcc502c43bcba66272d5ba` on 2026-03-15. The current onboarding defects on latest `alpha-test` are:

- `loongclaw onboard --non-interactive --skip-model-probe` is a dead end: onboarding warns that the model probe was skipped, then immediately blocks on that same warning even though the operator already accepted the skip explicitly.
- Explicit credential-source overrides can leave stale OAuth or inline provider secrets in place, so onboarding persists multiple credential forms instead of the one the operator selected.
- Non-interactive detected-starting-point selection does not follow the same sorted candidate order as the interactive starting-point picker, so a headless run can activate a different detected config than the interactive default.
- Interactive credential-source and system-prompt screens can show fallback copy that implies blank input will clear to inline/OAuth credentials or the built-in prompt, even though pressing Enter on a non-empty default reuses that prefilled value and there is no explicit clear affordance.
- Non-interactive onboarding still treats unresolved warnings as fatal even when the draft matches the existing on-disk config and onboarding would otherwise be a no-op, so operators cannot reuse a healthy existing config without forcing an unnecessary interactive review.
- Onboarding writes the target config before sqlite memory bootstrap completes. If bootstrap fails afterwards, the target config is left mutated instead of restoring the original file or removing the partial first-run output.

## Expected behavior

- Non-interactive onboarding should continue past only the warning that was explicitly accepted by `--skip-model-probe`, while still blocking any other unresolved warnings when a write or mutation would actually be needed.
- Choosing an API-key env source should clear superseded OAuth or inline provider secrets and persist only the selected credential form.
- Non-interactive detected-starting-point selection should match the same candidate ordering the interactive onboarding UI presents by default.
- Interactive onboarding should distinguish “use the default/prefilled value” from “clear back to fallback/built-in behavior”, and an explicit clear action should actually restore the built-in system prompt instead of silently keeping the custom prompt.
- If the existing config already matches the draft, non-interactive onboarding should report a successful no-op even when persistent warnings remain, because there is no new state to confirm or write.
- Config writes should be recoverable: if a post-write onboarding step fails, onboarding should restore the previous config or remove the partial new output.

## Reproduction steps

1. Check out `origin/alpha-test` at `66c1a99044f008acd0fcc502c43bcba66272d5ba`.
2. Run `loongclaw onboard --non-interactive --accept-risk --skip-model-probe` with a provider config that otherwise resolves cleanly, and observe that onboarding still aborts on the skipped model-probe warning.
3. Start from an existing config that already contains either `provider.oauth_access_token = "${OPENAI_CODEX_OAUTH_TOKEN}"` or an inline `provider.api_key`, rerun onboarding with `--non-interactive --api-key-env OPENAI_API_KEY --force`, and observe that the written config keeps the stale credential material instead of only the selected API-key env reference.
4. Provide multiple detected starting-point candidates in Codex config search paths whose sorted interactive order differs from raw discovery order, then compare the interactive default choice with the non-interactive selected output.
5. Start from an existing config with either `provider.api_key_env = "OPENAI_API_KEY"` or `cli.system_prompt = "custom review prompt"`, run interactive onboarding, reach the credential-source or system-prompt step, and observe that the screen copy implies blank input will fall back even though Enter reuses the non-empty default and there is no explicit clear action.
6. Start from an existing config whose effective onboarding draft is already identical to the saved file, rerun non-interactive onboarding with a warning-producing condition such as `--skip-model-probe`, and observe that onboarding still aborts instead of reporting that the existing config was kept.
7. Use an existing config, force an onboarding write, and point `memory.sqlite_path` at a path that passes directory preflight but fails sqlite bootstrap, then observe that onboarding errors after writing and leaves the mutated config on disk.

## Version / commit

`origin/alpha-test@66c1a99044f008acd0fcc502c43bcba66272d5ba`

## Contribution track impact

chumyin:
## Summary

- Allow non-interactive onboarding to accept the explicit `--skip-model-probe` warning while still blocking any other unresolved preflight warnings when a write is actually needed, and keep matching existing configs as successful no-op runs even if persistent warnings remain.
- Clear stale inline/OAuth provider credentials when onboarding switches to an API-key env source, and add an explicit `:clear` action on interactive onboarding screens so operators can intentionally fall back to inline/OAuth credentials or the built-in CLI system prompt when Enter would otherwise reuse a non-empty default.
- Align non-interactive detected-starting-point selection with the interactive sorted chooser, decouple accepted non-interactive warning handling from fragile display-text matching, and roll back config writes if sqlite bootstrap fails after the file write.
- This is needed because latest `alpha-test` still had headless onboarding dead ends, stale credential persistence, prompt-copy/behavior mismatches around non-empty defaults, no-op failure paths, and post-write recovery gaps.

## Scope

- [x] Small and focused
- [x] No unrelated refactors

## Risk Track

- The non-interactive warning gate now exempts only warnings explicitly tagged as accepted by `--skip-model-probe`; all other warnings still force an interactive review path when onboarding would otherwise mutate config state.
- Matching existing configs now short-circuit to the existing-config-kept no-op path even if persistent warnings remain, because onboarding is not introducing new risk in that branch.
- Interactive onboarding keeps Enter mapped to the shown default or prefilled value and introduces `:clear` as the explicit action for falling back to inline/OAuth credentials or the built-in prompt when the current default is non-empty.
- Credential-source overrides now clear superseded provider secrets before persistence, which changes onboarding behavior intentionally so the selected source is the only provider credential form that remains.
- System-prompt clear actions now restore the rendered built-in/native prompt instead of silently leaving the previous custom prompt in place.
- Config writes now keep a rollback recovery record until sqlite bootstrap succeeds, then clean up transient rollback backups on success.

## Validation

Additional validation notes:

- Added regressions for explicit `--skip-model-probe` acceptance, acceptance-policy decoupling from display text, non-interactive no-op reuse of matching existing configs under persistent warnings, clearing stale OAuth credentials, clearing stale inline API keys, interactive `:clear` fallback for credential-source and system-prompt prompts, non-interactive `:clear` fallback for those same knobs, matching interactive detected-starting-point ordering, first-write rollback cleanup, and restoring the original config when sqlite bootstrap fails after write.
- Before: `--skip-model-probe` still blocked non-interactive onboarding, warning acceptance depended on exact display strings, matching existing configs could still fail as non-interactive warning dead ends, `--api-key-env` could leave stale provider secrets behind, interactive fallback copy implied blank-input behavior that the prompt primitive could not actually perform with non-empty defaults, and post-write bootstrap failures could leave the config mutated.
- After: only explicitly accepted probe-skip warnings are bypassed, existing matching configs complete through the no-op kept-config path, the selected API-key env becomes the sole persisted provider credential form, interactive and non-interactive fallback actions use explicit `:clear` semantics where needed, non-interactive selection matches the interactive sorted default, and post-write bootstrap failures restore the previous or partial config state.
- Tests that mutate process-global env continue to serialize access through `DetectedEnvironmentGuard::without_detected_environment()` and the daemon test environment lock.

## Linked Issues

Closes #161

### DIFF
--- a/crates/daemon/src/onboard_cli.rs
+++ b/crates/daemon/src/onboard_cli.rs
@@ -11,6 +11,7 @@ use time::macros::format_description;
 
 const BACKUP_TIMESTAMP_FORMAT: &[FormatItem<'static>] =
     format_description!("[year][month][day]-[hour][minute][second]");
+const ONBOARD_CLEAR_INPUT_TOKEN: &str = ":clear";
 
 #[derive(Debug, Clone)]
 pub(crate) struct OnboardCommandOptions {
@@ -126,6 +127,10 @@ fn print_lines(ui: &mut impl OnboardUi, lines: impl IntoIterator<Item = String>)
 
 fn print_message(ui: &mut impl OnboardUi, line: impl Into<String>) -> CliResult<()> {
     ui.print_line(&line.into())
+}
+
+fn is_explicit_onboard_clear_input(raw: &str) -> bool {
+    raw.trim().eq_ignore_ascii_case(ONBOARD_CLEAR_INPUT_TOKEN)
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -272,6 +277,13 @@ impl ReviewFlowStyle {
     const fn header_subtitle(self) -> &'static str {
         crate::onboard_presentation::review_flow_copy(self.presentation_kind()).header_subtitle
     }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum SystemPromptSelection {
+    KeepCurrent,
+    RestoreBuiltIn,
+    Set(String),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -487,11 +499,9 @@ pub(crate) async fn run_onboard_cli_with_ui(
             resolve_api_key_env_selection(&options, &config, default_api_key_env, ui, context)?;
         apply_selected_api_key_env(&mut config.provider, selected_api_key_env);
 
-        if let Some(system_prompt) =
-            resolve_system_prompt_selection(&options, &config, ui, context)?
-        {
-            config.cli.system_prompt = system_prompt;
-        }
+        let system_prompt_selection =
+            resolve_system_prompt_selection(&options, &config, ui, context)?;
+        apply_selected_system_prompt(&mut config, system_prompt_selection);
     }
 
     let workspace_guidance = context
@@ -909,13 +919,16 @@ fn resolve_api_key_env_selection(
     context: &OnboardRuntimeContext,
 ) -> CliResult<String> {
     if options.non_interactive {
-        return Ok(options
-            .api_key_env
-            .as_deref()
-            .map(str::trim)
-            .filter(|value| !value.is_empty())
-            .unwrap_or(default_api_key_env.as_str())
-            .to_owned());
+        if let Some(api_key_env) = options.api_key_env.as_deref() {
+            if is_explicit_onboard_clear_input(api_key_env) {
+                return Ok(String::new());
+            }
+            let trimmed = api_key_env.trim();
+            if !trimmed.is_empty() {
+                return Ok(trimmed.to_owned());
+            }
+        }
+        return Ok(default_api_key_env);
     }
     let initial = options
         .api_key_env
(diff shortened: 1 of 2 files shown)


chumyin:
hi @gh-xj, i rebased this onto the latest head, fixed the non-interactive warning/no-op regression that showed up during review, and reran `cargo fmt --all -- --check`, `cargo clippy -j2 -p loongclaw-daemon --tests --no-deps -- -D warnings`, and `cargo test -j2 -p loongclaw-daemon`. the refreshed ci is green now, and branch protection still needs one approving review before merge.

PR #182, opened 2026-03-16
fix: harden provider active selection semantics

### LINKED ISSUE #175: [Bug] Prevent silent active-provider drift in mixed legacy/profile configs
## Summary
Mixed provider configs that combine legacy `[provider]` with saved `[providers.*]` profiles can silently change the effective active provider when `active_provider` is omitted.

## Problems
- loading a config with non-empty `providers` can ignore top-level `provider` semantics without warning
- active provider fallback currently depends on implicit profile ordering instead of an explicit selection
- `validate-config` does not report this semantic hazard
- `doctor`, `onboard`, and model probes only inspect the normalized active provider, which can mislead diagnosis when the active profile drifted
- model probing is treated as a hard readiness signal even for explicit-model configurations that do not require catalog discovery at runtime

## Expected behavior
- mixed legacy/profile configs should not silently change the active provider
- config validation should surface ambiguous or dangerous provider selection states
- diagnostics should clearly identify the active provider/profile being checked
- readiness flows should distinguish explicit-model chat viability from catalog-probe viability

## Scope
- provider runtime config normalization
- config validation diagnostics
- doctor/onboard provider diagnostics
- regression coverage for mixed config scenarios

chumyin:
## Summary

- preserve explicit legacy `[provider]` intent during TOML parsing and normalization so mixed configs no longer drift to the first saved profile when `active_provider` is missing or invalid
- add severity-aware config diagnostics and `validate-config` summaries so warning-only provider-selection recovery is visible without being treated as a hard config failure
- downgrade explicit-model catalog probe failures in `doctor` and `onboard` to warnings and include active profile context in provider diagnostics

## Scope

- [x] Small and focused
- [x] No unrelated refactors

## Risk Track

- Changes provider-selection fallback for mixed legacy/profile configs by preserving raw TOML selection intent before normalization.
- Warning-only validation is now reported as `valid=true` with separate `error_count` and `warning_count`; hard errors still fail semantic validation.
- Added regression coverage for the explicit legacy path, invalid or implicit selection recovery, conflicting recovered profile IDs, and explicit-model `/models` probe failures.

## Validation

Before/after behavior and regression coverage:

- Before: mixed `[provider]` plus `[providers.*]` configs could silently activate the first normalized profile, and invalid or missing `active_provider` recovery was effectively ordered-map dependent.
- After: valid explicit `active_provider` still wins; otherwise explicit legacy `[provider]` intent is matched or recovered deterministically, warning diagnostics are emitted for implicit or unknown active-provider recovery, and explicit-model catalog probe failures stay advisory in `doctor` and `onboard`.
- Fresh verification run:
  - `cargo test -p loongclaw-app -- --test-threads=1`
  - `cargo test -p loongclaw-daemon -- --test-threads=1`
- Targeted regression coverage added for:
  - mixed configs preserving explicit legacy provider when `active_provider` is absent
  - warning diagnostics for implicit or unknown active-provider recovery
  - legacy-provider field validation even when saved profiles exist
  - conflicting recovered profile IDs
  - explicit-model probe warnings in `doctor` and `onboard`
  - warning-only `validate-config` summaries and active-profile-aware credential or probe details
- Process-global env mutations in onboarding tests are serialized and restored with `DetectedEnvironmentGuard`.
- Follow-up commit `688c0a0` removed the `clippy::unreachable` violation in provider-profile recovery; the rerun CI for this PR head is fully green.

## Linked Issues

Closes #175

### DIFF
--- a/crates/app/src/config/channels.rs
+++ b/crates/app/src/config/channels.rs
@@ -1204,6 +1204,7 @@ fn validate_channel_account_integrity<'a, I>(
         );
         extra_message_variables.insert("raw_account_labels".to_owned(), labels.join(", "));
         issues.push(ConfigValidationIssue {
+            severity: super::shared::ConfigValidationSeverity::Error,
             code: ConfigValidationCode::DuplicateChannelAccountId,
             field_path: format!("{channel_key}.accounts"),
             inline_field_path: format!("{channel_key}.accounts.{normalized_account_id}"),
@@ -1242,6 +1243,7 @@ fn validate_channel_account_integrity<'a, I>(
             .join(", "),
     );
     issues.push(ConfigValidationIssue {
+        severity: super::shared::ConfigValidationSeverity::Error,
         code: ConfigValidationCode::UnknownChannelDefaultAccount,
         field_path: format!("{channel_key}.default_account"),
         inline_field_path: format!("{channel_key}.accounts"),
--- a/crates/app/src/config/runtime.rs
+++ b/crates/app/src/config/runtime.rs
@@ -15,8 +15,8 @@ use super::{
     feishu_integration::FeishuIntegrationConfig,
     provider::{ProviderConfig, ProviderKind, ProviderProfileConfig},
     shared::{
-        ConfigValidationIssue, ConfigValidationLocale, DEFAULT_CONFIG_FILE,
-        default_loongclaw_home as shared_default_loongclaw_home, expand_path,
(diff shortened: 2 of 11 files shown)


chumyin:
Follow-up status:

- pushed `688c0a0` to remove the `clippy::unreachable` failure in provider-profile recovery
- reran CI on the updated PR head and all current checks are green
- no open review threads or inline comments remain

Ready for review.

PR #260, opened 2026-03-17
fix(onboard): numbered selection and stdin drain

gh-xj:
## Summary

- Replace `[slug]` bracket display with numbered selection (`1)`, `2)`, `3)`) for personality, memory profile, and provider prompts — users no longer need to type exact slug strings
- Add `drain_stdin()` after each `read_line()` using Unix non-blocking I/O to discard buffered bytes, preventing multi-line paste from corrupting subsequent prompts
- Extend `OnboardUi` trait with `select_one()` method (`Option<usize>` default — `None` forces explicit choice for ambiguous provider selection)

## Scope

- [x] Small and focused
- [x] Includes docs updates (if needed)
- [x] No unrelated refactors

## Risk Track

- [x] Track A (routine/low-risk)
- [ ] Track B (higher-risk/policy-impacting)

## Validation

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-features` (517 daemon tests pass)
- [ ] Manual: personality selection shows numbered list, accepts 1/2/3
- [ ] Manual: memory profile selection shows numbered list
- [ ] Manual: paste multi-line text — subsequent prompts not corrupted
- [ ] Manual: `--non-interactive` mode unchanged
- [ ] Manual: provider selection with forced choice (no default)

## Linked Issues

Closes #247
Closes #248

## Summary by CodeRabbit

* **New Features**
  * Onboarding UI: numbered selection prompts with labeled options, descriptions, “recommended” markers, defaults, clearer header screens, and consistent mapping of choices.

* **Bug Fixes**
  * Terminal input handling: stdin is drained on prompt exit to prevent leftover or pasted input from affecting later prompts.

* **Tests**
  * Integration tests updated to cover numeric selection flows, defaults, cancellation, and non-interactive paths.

Issue #247, opened 2026-03-17
[Bug]: Multiline paste in onboard prompt corrupts subsequent input steps


gh-xj:
### Area

Migration / onboarding

### Severity

S1 - workflow blocked

### Regression?

No, this is a first-time setup or newly exercised path

### Summary

`StdioOnboardUi::prompt_with_default` and `prompt_required` use `io::stdin().read_line()` which captures exactly one line. When a user pastes multiline text (e.g., at the "Prompt addendum" or "CLI system prompt" step), only the first line is consumed. Remaining lines stay in the stdin buffer and are silently consumed by subsequent `read_line()` calls, corrupting downstream prompts (personality, memory profile, review) and causing unexpected validation errors or flow termination.

### Expected behavior

Either accept multiline input safely, or drain the stdin buffer after each `read_line()` call so leftover lines do not corrupt subsequent prompts. At minimum, display a hint that the field is single-line only.

### Reproduction steps

```bash
1. loongclaw onboard --force
2. Accept defaults for Provider, Model, Credential steps (press Enter)
3. At step 5 "Prompt addendum", paste two lines:
   You are helpful.
   Always be concise.
4. Observe the second line is consumed by the next prompt (Memory profile),
   causing "Unsupported memory profile: Always be concise" or similar cascading errors.
```

### Version / commit

alpha-test@6709727

### Operating system

macOS 26

### Runtime context

terminal=Ghostty 72x24, interactive onboard flow

### Config / environment context

Default config, fresh onboarding via `--force`

### Logs / screenshots / evidence

Root cause in `crates/daemon/src/onboard_cli.rs`:
- Line 83: `read_line()` in `prompt_with_default`
- Line 99: `read_line()` in `prompt_required`
- Line 112: `read_line()` in `prompt_confirm`

### Impact

Affected users: any new user who pastes multiline text during onboarding
Frequency: every time multiline text is pasted
Consequence: corrupted config flow, potential bad config written silently

### Pre-flight checks

- [x] I reproduced this on the latest `alpha-test` branch tip or the latest published release.
- [x] I redacted secrets, tokens, and private endpoints from everything I attached.
- [x] I searched existing issues and did not find the same report already tracked.

chumyin:
fixed in #260. onboarding now drains leftover stdin after prompt reads and keeps the flow stable when multiline input is pasted, so the follow-up corruption path is resolved on alpha-test.

chumyin:
[...]

PR #273, opened 2026-03-17
fix(onboard): harden single-line prompt capture

chumyin:
## summary

- replace the unix-specific prompt drain with queued single-line capture inside `StdioOnboardUi`
- drop follow-up pasted lines at the prompt boundary instead of letting them leak into later onboarding steps
- surface single-line guidance on the prompt addendum and system prompt screens

## root cause

`#260` stopped the immediate corruption path, but it still solved the symptom by draining stdin after `read_line()` with unix-specific fd handling. the deeper issue was that the stdio onboarding ui did not model prompt ownership for extra pasted lines. this follow-up moves that boundary into the ui itself.

## validation

- `cargo test --workspace`
- `LOONGCLAW_ARCH_STRICT=true bash scripts/check_architecture_boundaries.sh`
- `bash scripts/check_dep_graph.sh`
- `bash scripts/check-docs.sh` (passes with existing release-artifact warnings)
- `cargo deny check` (passes with existing duplicate-dependency and deny config warnings)

## notes

- `task verify` is not available in this environment because `task` is not installed
- `task check:conventions` remains blocked here because `~/.claude/skills/convention-engineering/scripts/main.go` is unavailable

## linked issues

- refs #247
- follow-up to #260

## Summary by CodeRabbit

* **Improvements**
  * Enhanced handling of multi-line pastes for single-line onboarding prompts to avoid accidental input.
  * Added on-screen hints informing users that input is single-line and extra pasted lines are ignored.
  * Emits a clear notice when extra pasted lines are dropped to prevent confusion.

* **Tests**
  * Added coverage for single-line prompt capture, paste-drain behavior, and presence of the single-line input hint.

### DIFF
--- a/crates/daemon/src/onboard_cli.rs
+++ b/crates/daemon/src/onboard_cli.rs
@@ -18,7 +18,8 @@ const ONBOARD_CLEAR_INPUT_TOKEN: &str = ":clear";
 const ONBOARD_ESCAPE_CANCEL_HINT: &str = "- press Esc then Enter to cancel onboarding";
 const ONBOARD_SINGLE_LINE_INPUT_HINT: &str =
     "- terminal input is single line; extra pasted lines are ignored";
-const ONBOARD_PASTE_DRAIN_WINDOW: Duration = Duration::from_millis(20);
+const ONBOARD_PASTE_DRAIN_WINDOW_ENV: &str = "LOONGCLAW_ONBOARD_PASTE_DRAIN_WINDOW_MS";
+const DEFAULT_ONBOARD_PASTE_DRAIN_WINDOW: Duration = Duration::from_millis(75);
 const ONBOARD_LINE_READER_BUFFER_SIZE: usize = 64;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


chumyin:
addressed the follow-up review items on this branch:

- bounded the onboarding stdin queue so multi-line paste applies backpressure instead of unbounded growth
- kept the direct spawn-failure fallback explicit with a one-time degraded-mode warning
- replaced the fixed paste drain timeout with a configurable per-session window via `LOONGCLAW_ONBOARD_PASTE_DRAIN_WINDOW_MS` (default `75ms`) so slower terminals do not leak delayed paste chunks into the next prompt
- added regression coverage for queue backpressure, degraded fallback behavior, and drain-window override fallback parsing

verification on this branch:

- `cargo clippy -p loongclaw-daemon --all-targets --all-features -- -D warnings`
- `cargo test -p loongclaw-daemon onboard_paste_drain_window_ -- --nocapture`
- `cargo test -p loongclaw-daemon onboard_cli -- --nocapture` (the same 5 existing local integration failures still stop at `TcpListener::bind("127.0.0.1:0")` with `Operation not permitted` in this environment)

refreshed github checks on `566ed44` are green.

PR #284, opened 2026-03-18
feat: retire implicit noop audit constructor default

### LINKED ISSUE #279: [Feature] Retire the implicit NoopAuditSink kernel constructor default
## Area

Kernel / audit / runtime safety

## Problem statement

`alpha-test` now has durable audit retention on the main app/bootstrap path, and
built-in audit inspection is being finished in PR #259. However, one quieter
footgun still remains in the kernel layer:

- `LoongClawKernel::new()` still defaults to `NoopAuditSink`
- multiple kernel tests and helper paths still rely on that implicit constructor
- this means a caller can instantiate a kernel that silently drops security-
  critical audit events unless they remember to choose `with_runtime(...)`

That leaves an architecture truth gap. The repository is moving toward a
kernel-first governance story where policy decisions and runtime-plane events
should always have an evidence path, but one of the easiest constructors still
creates a silent-drop sink by default.

This is not the same problem as durable app bootstrap anymore; that slice is
already landed or in flight. This is the follow-up constructor-safety problem
that becomes more visible now that durable audit is otherwise treated as the
safe default.

## Proposed solution

Retire the implicit `NoopAuditSink` default from general-purpose kernel
construction.

Recommended shape:

- make the audit choice explicit at the constructor boundary for non-test code
- keep a deliberately named test-only/no-audit helper if some fixtures still
  need side-effect-free kernels
- update kernel/app test helpers so intentional no-audit usage is explicit and
  grep-able
- add regression coverage proving the default/general constructor no longer
  silently drops audit events
- update reliability/docs wording so it matches the real constructor behavior

## Non-goals / out of scope

- reworking the durable JSONL/fanout app bootstrap that already exists
- adding indexed audit storage, retention rotation, or remote export
- redesigning audit event schemas or trace identifiers
- broad approval/runtime closure work outside this constructor seam

## Alternatives considered

1. Leave `NoopAuditSink` in place and rely on contributor discipline.
   - Rejected because this is exactly the kind of hidden default that creates
     false confidence in safety-sensitive code paths.
2. Switch every constructor directly to durable JSONL.
   - Rejected because many unit/integration fixtures still need explicit local
     control over side effects.
3. Only patch docs to mention the exception.
   - Rejected because the root problem is architectural behavior, not wording.

## Acceptance criteria

- the general-purpose kernel constructor path no longer silently drops audit
  events by default
- any remaining no-audit constructor is explicitly named and scoped for tests or
  narrow fixture use
- regression tests prove the new constructor behavior and cover the intentional
  no-audit escape hatch
- reliability/security docs stop overclaiming constructor-level audit safety
- the change is reviewed as a bounded follow-up to #172 / PR #259 rather than
  folded into unrelated runtime hardening work

## Policy / security / breaking sensitivity

Yes, this touches policy, security, or potentially breaking behavior

## Rollout / rollback notes

Roll out on `alpha-test` first. If fixture churn is larger than expected, keep a
clearly named test-only no-audit helper as an intermediate compatibility seam,
but do not keep the silent general-purpose default.

## Impact

Affected users: maintainers and contributors who instantiate kernels directly
Frequency: every new test/helper/runtime path that reaches for the default
constructor
Consequence if missing: silent audit drops remain possible in paths that look
safe by default, weakening the credibility of the runtime evidence boundary

## Related context

- Follow-up to #172
- Related to #263
- Adjacent to PR #259

chumyin:
## Summary

- Problem: the default kernel constructor still wired `NoopAuditSink` implicitly, so repository-owned tests and side-effect-free local runs could miss observable audit behavior unless every caller remembered to opt into a real sink.
- Why it matters: the runtime and test surface should make audit behavior explicit and observable by default instead of silently discarding events through an implicit noop path.
- What changed: switched the default kernel constructor to in-memory audit, added explicit constructor helpers for in-memory and no-audit fixture paths, migrated repository-owned tests to the explicit in-memory helper, added regression coverage, and updated reliability/security docs plus design docs for the default-behavior change.
- What did not change (scope boundary): this does not add durable retention, change daemon CLI audit commands, or alter unrelated provider/runtime policy.

## Linked Issues

- Closes #279
- Related # n/a

## Change Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [x] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [x] Kernel / policy / approvals
- [x] Contracts / protocol / spec
- [ ] Daemon / CLI / install
- [ ] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [ ] Channels / integrations
- [x] Docs / contributor workflow

## Risk Track

If Track B, fill these in:

- Risk notes:
  this changes the default audit behavior of the kernel bootstrap path, so callers and tests that previously relied on implicit noop behavior must remain explicit and well covered.
- Rollout / guardrails:
  explicit in-memory and no-audit helpers were added for fixture paths, repository-owned tests were migrated, and regression coverage verifies the new default behavior.
- Rollback path:
  revert the default-constructor change to restore implicit noop audit behavior.

## Validation

- [ ] `cargo test --workspace --locked`
- [ ] `cargo test --workspace --all-features --locked`
- [x] Relevant architecture / dep-graph / docs checks for touched areas

Commands and evidence:

```text
cargo fmt --all -- --check
- passed

cargo test --workspace -- --test-threads=1
- passed

cargo test --workspace --all-features -- --test-threads=1
- passed

- passed

./scripts/check_architecture_boundaries.sh
- passed

./scripts/check_dep_graph.sh
- passed

./scripts/check-docs.sh
- passed
```

Before / after behavior:

- Before: the default kernel constructor implicitly wired `NoopAuditSink`, so callers that wanted observable audit behavior had to opt in explicitly.
- After: the default kernel constructor uses in-memory audit, while explicit helpers remain available for in-memory and no-audit fixture paths.

Regression coverage:

- repository-owned tests were moved onto the explicit in-memory helper where appropriate.
- kernel and spec tests assert the new default constructor behavior and the retained explicit no-audit path.

- kernel bootstraps now observe audit events in memory by default instead of implicitly dropping them through `NoopAuditSink`.

## Failure Recovery

- Fast rollback or disable path: revert this change to restore the previous implicit noop default.
- Observable failure symptoms reviewers should watch for: callers unexpectedly depending on missing audit events, tests still assuming the implicit noop path, or fixture paths no longer selecting the intended audit mode explicitly.

## Reviewer Focus

- `crates/kernel/src/kernel.rs` for the default-constructor behavior and explicit helper split.
- `crates/kernel/src/tests.rs`, `crates/kernel/tests/kernel_integration.rs`, and spec tests for regression coverage around the default audit path.
- `docs/RELIABILITY.md` and `docs/SECURITY.md` for the documented default-behavior change.

### DIFF
--- a/crates/app/src/memory/tests.rs
+++ b/crates/app/src/memory/tests.rs
@@ -29,7 +29,8 @@ async fn mvp_memory_adapter_routes_through_kernel() {
         ExecutionRoute, HarnessKind, LoongClawKernel, StaticPolicyEngine, VerticalPackManifest,
     };
 
-    let mut kernel = LoongClawKernel::new(StaticPolicyEngine::default());
+    let (mut kernel, _audit) =
+        LoongClawKernel::new_with_in_memory_audit(StaticPolicyEngine::default());
 
     kernel.register_core_memory_adapter(MvpMemoryAdapter::new());
     kernel
--- a/crates/kernel/src/kernel.rs
+++ b/crates/kernel/src/kernel.rs
@@ -8,7 +8,10 @@ use std::{
 };
 
 use crate::{
-    audit::{AuditEvent, AuditEventKind, AuditSink, ExecutionPlane, NoopAuditSink, PlaneTier},
+    audit::{
+        AuditEvent, AuditEventKind, AuditSink, ExecutionPlane, InMemoryAuditSink, NoopAuditSink,
+        PlaneTier,
+    },
     clock::{Clock, SystemClock},
     connector::{ConnectorExtensionAdapter, ConnectorPlane, CoreConnectorAdapter},
     contracts::{
@@ -84,8 +87,29 @@ pub struct Kernel<P: PolicyEngine> {
 }
 
 impl<P: PolicyEngine> LoongClawKernel<P> {
+    /// Safe convenience constructor for callers that do not need to customize
+    /// runtime components. This defaults to in-memory audit rather than silent
+    /// audit dropping.
     #[must_use]
     pub fn new(policy: P) -> Self {
+        Self::new_with_in_memory_audit(policy).0
+    }
+
+    /// Construct a kernel with the default system clock and an inspectable
+    /// in-memory audit sink.
+    #[must_use]
+    pub fn new_with_in_memory_audit(policy: P) -> (Self, Arc<InMemoryAuditSink>) {
+        let audit = Arc::new(InMemoryAuditSink::default());
+        let kernel = Self::with_runtime(policy, Arc::new(SystemClock), audit.clone());
+        (kernel, audit)
+    }
+
+    /// Construct a kernel that intentionally discards audit events.
+    ///
+    /// This is reserved for narrow fixture paths where callers explicitly do
+    /// not need audit assertions or evidence retention.
+    #[must_use]
+    pub fn new_without_audit(policy: P) -> Self {
         Self::with_runtime(policy, Arc::new(SystemClock), Arc::new(NoopAuditSink))
     }
 
--- a/crates/kernel/src/tests.rs
+++ b/crates/kernel/src/tests.rs
@@ -105,6 +105,37 @@ fn fanout_audit_sink_records_to_all_children() {
 }
 
 #[test]
+fn explicit_in_memory_kernel_constructor_records_token_audit_events() {
+    let (mut kernel, audit) =
+        LoongClawKernel::new_with_in_memory_audit(StaticPolicyEngine::default());
+    kernel
+        .register_pack(sample_pack())
+        .expect("pack should register");
+
+    kernel
+        .issue_token("sales-intel", "agent-in-memory", 120)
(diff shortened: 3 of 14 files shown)


chumyin:
rebased this onto current `dev`, fixed the implementation-plan heading-level nit, and reran the local verification path here:

- `cargo test -p loongclaw-spec --test spec_execution`
- `cargo test -p loongclaw-spec kernel_bootstrap --lib`
- `cargo test -p loongclaw-kernel explicit_ -- --test-threads=1`
- `cargo test --workspace -- --test-threads=1`
- `cargo test --workspace --all-features -- --test-threads=1`
- `./scripts/check_architecture_boundaries.sh`
- `./scripts/check_dep_graph.sh`
- `./scripts/check-docs.sh`

refreshed github checks are green on `c5c037a7`; the remaining blocker is the required approving review before merge.

PR #336, opened 2026-03-18
feat: converge spec audit contract on named in-memory default

### LINKED ISSUE #335: Explicitly name and test the spec in-memory audit default
## Summary

The `#279` constructor fix made `LoongClawKernel::new()` safe by default, but the `spec` layer still leaves its in-memory audit contract implicit.

Today:
- `crates/spec/src/kernel_bootstrap.rs` falls back to `InMemoryAuditSink` without a named harness-level contract
- `crates/spec/src/spec_execution.rs` allocates its own in-memory sink inline instead of reusing a shared spec bootstrap decision
- docs already allow `spec`/demo helpers to remain in-memory, but the behavior is not protected by focused regression coverage

That leaves a semantic gap between production-shaped audit defaults and harness-shaped spec execution.

## Why This Matters

We do not want to pull `spec` into the app runtime audit configuration surface just to fix this gap. The smaller and cleaner change is to keep `spec` intentionally in-memory, but make that choice explicit, centralized, and testable.

## Acceptance Criteria

- introduce a named spec/bootstrap helper for the default in-memory audit sink
- route spec bootstrap fallback and spec execution through that helper
- add regression tests proving spec execution can retain audit snapshots when requested and suppress them when not requested
- add a focused bootstrap test proving the named helper records real audit events
- align `docs/SECURITY.md` and `docs/RELIABILITY.md` with the resulting contract

## Context

This is a follow-up to `#279` and fits under the broader runtime hardening work tracked in `#263` and `#290`.

chumyin:
## Summary

- add a named spec/bootstrap helper for the default in-memory audit sink
- route spec bootstrap fallback and spec execution through that helper
- add regression coverage for `include_audit = true|false` and a focused bootstrap audit test
- tighten security/reliability docs around the production durable vs. spec in-memory audit split

## Why

`#279` fixed the unsafe implicit noop audit default at the kernel constructor level, but the `spec` layer still left its in-memory audit semantics implicit. This change keeps `spec` harness-shaped and side-effect-free while making that audit contract explicit, centralized, and regression-tested.

## Validation

- `cargo test --workspace -- --test-threads=1`
- `cargo test --workspace --all-features -- --test-threads=1`
- `./scripts/check_architecture_boundaries.sh`
- `./scripts/check_dep_graph.sh`
- `./scripts/check-docs.sh`

## Notes

- This PR is intentionally stacked on `#284`; after `#284` lands it can be retargeted to `alpha-test`.

Closes #335

### DIFF
--- a/crates/spec/src/kernel_bootstrap.rs
+++ b/crates/spec/src/kernel_bootstrap.rs
@@ -17,7 +17,7 @@ use crate::spec_runtime::{
 
 /// The spec/bootstrap layer is a harness-facing surface, so its default audit
 /// sink stays explicitly in-memory unless a caller wires a different sink.
-pub fn default_in_memory_audit_sink() -> Arc<InMemoryAuditSink> {
+pub(crate) fn default_in_memory_audit_sink() -> Arc<InMemoryAuditSink> {
     Arc::new(InMemoryAuditSink::default())
 }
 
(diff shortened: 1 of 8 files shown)


chumyin:
hello, i tightened this stacked slice a bit before merge: added real coverage for the default and clock-only bootstrap fallback paths, reduced the helper to crate-only visibility, and refreshed the march architecture drift report so governance stays green. local verification plus the full pr ci are green, so i'm rebasing this into #284 now.

Issue #304, opened 2026-03-18
[P1] Cross-platform CI matrix (macOS + Windows)


gh-xj:
## Summary
CI only tests on Linux despite shipping macOS and Windows binaries. Add cross-platform test matrix.

**Epic**: #294
**Priority**: P1
**Effort**: M
**Week**: 3-4

## Acceptance Criteria
- [ ] CI matrix includes: ubuntu-latest, macos-latest, windows-latest
- [ ] `cargo test --workspace` passes on all 3 platforms
- [ ] `cargo build` passes on all 3 platforms
- [ ] Platform-specific test skips documented where needed

## How to Claim
Comment with your GitHub handle to claim this issue.

## References
- Epic: #294
- Related: PR #50 (Windows stack size fix, merged)

Mike-7777777:
## Windows Test Report — `cargo test` on Windows 11

**Environment**: Windows 11 Pro 10.0.26200, Rust 1.93.1, branch `dev` @ `8e63e04`

### Results

```
test result: FAILED. 2029 passed; 56 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.61s
```

All 56 failures are Windows-specific — they fall into two root causes.

---

### Root Cause 1: Path separator (`\` vs `/`)

**~10 failures** across 3 modules. Tests assert paths with hardcoded `/`, but Windows APIs return `\`.

| Module | Example assertion | File |
|--------|-------------------|------|
| `memory::workspace_files` (2) | `left: "workspace\MEMORY.md"` / `right: "workspace/MEMORY.md"` | `workspace_files.rs:292` |
| `memory::durable_recall` (2) | `left: "memory\2026-03-22.md"` / `right: "memory/2026-03-22.md"` | `durable_recall.rs:202, 220` |
| `config::audit` (1) | `config.path.ends_with(".loongclaw/audit/events.jsonl")` fails on `\` path | `audit.rs:78` |
| `tools` (feishu resource) (4) | `saved_path.ends_with("/artifacts/audio/voice.ogg")` fails on `\` path | `tools/mod.rs:7796, 7970, 8141, 8308` |

**Fix approach**: Normalize paths before comparison, or use `std::path::Path` methods instead of string `ends_with`.

---

### Root Cause 2: File locking semantics

**~46 failures** across 2 modules. Windows does not allow deleting files that are open by another handle.

#### 2a. SQLite test lock poisoning (~43 failures)

```
[...]

PR #585, opened 2026-03-26
fix(windows): normalize path separators in display labels and test assertions

Mike-7777777:
## Summary

Fixes 9 of 56 Windows test failures tracked in #304.

- **`workspace_memory_label`**: Normalize `\` → `/` in the returned label string. These are display labels shown to the LLM/user, not filesystem paths — forward slash is universally readable.
- **Audit config test**: Use `std::path::Path::join` for platform-agnostic suffix matching instead of hardcoded `/` string.
- **Feishu tool tests**: Use `std::path::Path::ends_with` instead of `str::ends_with` with hardcoded `/` paths.

## Test Results (Windows 11 Pro, Rust 1.93.1)

**Before:** 9 failures in `workspace_files`, `durable_recall`, `audit`, and `tools` modules.
**After:** All 9 pass. No regressions — 2085 total tests pass.

## Files Changed

| File | Change |
|------|--------|
| `crates/app/src/memory/workspace_files.rs` | 1 line: add `.replace('\', "/")` to label output |
| `crates/app/src/config/audit.rs` | Test-only: platform-agnostic path assertion |
| `crates/app/src/tools/mod.rs` | Test-only: 4 assertions use `Path::ends_with` |

## Checklist

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes
- [x] `cargo test -p loongclaw-app --lib` passes (2085/2085)
- [x] No new dependencies added

Part 1 of 3 for #304 (cross-platform CI). See also the companion PRs for SQLite UNC paths and Feishu runtime file locking.

## Summary by CodeRabbit

* **Tests**
  * Improved audit configuration testing with platform-aware path comparisons.
  * Enhanced file location verification in Feishu message tests with OS-appropriate path assertions.

* **Bug Fixes**
  * Normalized path separators in workspace labels for consistent cross-platform formatting.

PR #586, opened 2026-03-26
fix(windows): use dunce::canonicalize to avoid UNC path mismatch in SQLite runtime

Mike-7777777:
## Summary

Fixes ~43 of 56 Windows test failures tracked in #304. All in `memory::sqlite::tests`.

**Root cause:** `std::fs::canonicalize()` on Windows returns `\?\C:\...` UNC extended-length paths. When comparing these against non-UNC paths via `PathBuf::starts_with()`, the comparison always fails because `Component::Prefix` differs (`VerbatimDisk` vs `Disk`).

**Fix:** Replace `std::fs::canonicalize` with `dunce::canonicalize` in the SQLite runtime path normalization. `dunce` strips the `\?\` prefix when the path doesn't require it (< 260 chars, no special characters), which is the standard Rust solution for this Windows UNC issue.

**Bonus fix:** The shared test `Mutex` used `.expect("runtime test lock")`, which causes one test failure to cascade into ~42 `PoisonError` panics. Changed to `.unwrap_or_else(|poisoned| poisoned.into_inner())` so each test can proceed independently.

## Test Results (Windows 11 Pro, Rust 1.93.1)

**Before:** 43 failures in `memory::sqlite::tests` (1 real + 42 cascade from lock poisoning).
**After:** All 55 SQLite tests pass. No regressions — 2085 total tests pass.

## Files Changed

| File | Change |
|------|--------|
| `crates/app/Cargo.toml` | Add `dunce = "1"` dependency |
| `Cargo.lock` | Updated lockfile |
| `crates/app/src/memory/sqlite.rs` | Replace 2 `canonicalize()` calls with `dunce::canonicalize()`; replace 44 `.expect("runtime test lock")` with poison-resilient `.unwrap_or_else()` |

## Note on broader canonicalize usage

There are ~30 other `.canonicalize()` calls across `crates/app/src/` (in `migration/`, `tools/`, etc.). These were not changed in this PR because they don't cause test failures — the UNC mismatch specifically affects path comparison logic in the SQLite runtime cache. A follow-up PR could address the remaining calls for consistency.

## Checklist

- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` passes
- [x] `cargo test -p loongclaw-app --lib` passes (2085/2085)
- [x] New dependency `dunce` is already in `Cargo.lock` as transitive dep; now direct

Part 2 of 3 for #304 (cross-platform CI). See also companion PRs for path separators and Feishu runtime file locking.

## Summary by CodeRabbit

* **Chores**
  * Added new dependency to enhance file path handling capabilities across different operating systems.

* **Bug Fixes**
  * Improved database path normalization to more reliably handle edge cases in path canonicalization.
  * Enhanced test suite resilience by allowing test lock mechanisms to gracefully recover from poisoned states, reducing potential intermittent test failures.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2002,6 +2002,7 @@ dependencies = [
  "cbc",
  "chrono",
  "console",
+ "dunce",
  "flate2",
  "futures-util",
  "hex",
--- a/crates/app/Cargo.toml
+++ b/crates/app/Cargo.toml
@@ -31,6 +31,7 @@ loongclaw-contracts = { path = "../contracts" }
 loongclaw-kernel = { path = "../kernel" }
 async-trait.workspace = true
 console.workspace = true
+dunce = "1"
 rand.workspace = true
 chrono.workspace = true
 serde.workspace = true
--- a/crates/app/src/memory/sqlite.rs
+++ b/crates/app/src/memory/sqlite.rs
@@ -828,8 +828,7 @@ fn normalize_runtime_db_path(path: &Path) -> Result<PathBuf, String> {
     test_support::record_runtime_path_normalization_full();
 
     let normalized = if absolute.exists() {
-        absolute
-            .canonicalize()
+        dunce::canonicalize(&absolute)
             .map_err(|error| format!("canonicalize sqlite db path failed: {error}"))?
     } else {
         let Some(file_name) = absolute.file_name() else {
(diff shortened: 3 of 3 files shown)


chumyin:
hi @Mike-7777777, thanks for splitting the windows follow-up into focused fixes. i re-checked the sqlite runtime path normalization flow, reviewed the companion context in #304 / #585 / #587, and reran the affected sqlite coverage locally. `dunce::canonicalize` is the right fix for the windows verbatim-path mismatch here, and the poisoned-lock recovery makes the test surface a lot easier to reason about. nice work on this one, merging it now.

PR #590, opened 2026-03-26
ci: add Windows to test matrix

Mike-7777777:
## Summary

Add `windows-latest` to the `rust-test-default` and `rust-test-all-features` CI matrix, with `fail-fast: false` so each platform reports independently.

## CI Status

| Platform | rust-test-default | rust-test-all-features |
|----------|-------------------|------------------------|
| Ubuntu | ✅ pass | ✅ pass |
| Windows | ❌ 7 daemon integration test failures (#610) | ❌ same 7 failures |

The 7 failures are **pre-existing Windows compatibility issues** (8.3 short names + UNC paths in daemon integration tests), not caused by this CI config change. They are tracked in #610.

The original 2 sqlite path normalization failures (#592) were fixed in #593 (merged). These 7 are new discoveries in the daemon crate — the exact reason we need Windows CI.

## Changes

- `.github/workflows/ci.yml`: add `strategy.matrix.os: [ubuntu-latest, windows-latest]` with `fail-fast: false` to `rust-test-default` and `rust-test-all-features`
- Linux-only jobs unchanged: `governance`, `rust-quality`, `docs-build`, `android-termux-build`

## macOS

Deferred to a separate PR pending #591 (pre-existing flaky SQLite WAL locking test).

## Related

- Tracked in #304
- Windows daemon test failures: #610
- macOS flaky test: #591

## Summary by CodeRabbit

* **Chores**
  * Expanded test coverage to run on both Ubuntu and Windows operating systems, improving cross-platform validation of the codebase.

PR #593, opened 2026-03-26
fix(windows): resolve 8.3 short names in sqlite path normalization

### LINKED ISSUE #592: [Bug] Windows CI: SQLite path normalization tests fail on GitHub Actions runner
## Description

Two SQLite path normalization tests pass locally on Windows 11 but fail on the `windows-latest` GitHub Actions runner:

- `memory::sqlite::tests::dot_dot_aliases_share_one_runtime_after_normalization`
- `memory::sqlite::tests::equivalent_relative_and_absolute_paths_share_one_runtime`

Both assert `left: 0, right: 1` — the bootstrap count under the temp-dir prefix returns 0.

**Fix: #593** — adds `canonicalize_existing_ancestor()` to resolve 8.3 short names in path normalization.

## Root Cause

On GitHub Actions `windows-latest`, `std::env::temp_dir()` returns a path with 8.3 short names (e.g. `C:\Users\RUNNER~1\...`). When `normalize_runtime_db_path` encounters a DB path whose parent directory does not yet exist, `dunce::canonicalize(parent)` fails and falls back to the raw path — preserving the short name. Later lookups canonicalize the same prefix to the long name (`runneradmin`), causing `PathBuf::starts_with` mismatches.

## Environment Comparison

| | Local (passes) | CI (fails) |
|---|---|---|
| OS | Windows 11 Pro | Windows Server 2022 |
| temp_dir | `C:\Users\<longname>\...\Temp` | `C:\Users\RUNNER~1\...\Temp` (8.3 short name) |
| Drive | `I:\` | `D:\a\` |

## Related

- Part of #304 (Windows test compatibility)
- Discovered via #590 (cross-platform CI matrix)
- Prior fix #586 (`dunce::canonicalize`) handles UNC paths but not 8.3 short names when parent doesn't exist

Mike-7777777:
## Summary

Fixes #592 — two SQLite path normalization tests fail on `windows-latest` CI runner but pass locally.

**Root cause:** On GitHub Actions `windows-latest`, `std::env::temp_dir()` returns a path containing 8.3 short names (e.g. `C:\Users\RUNNER~1\...`). When `normalize_runtime_db_path` encounters a DB path whose parent directory does not yet exist, `dunce::canonicalize(parent)` fails and the code falls back to `absolute.clone()`, preserving the short name. Later lookups that canonicalize the same prefix resolve to the long name (`runneradmin`), causing `PathBuf::starts_with` mismatches.

This is also a latent production bug: if a user's working directory or config path contains 8.3 short names, the first DB access (before parent dirs exist) stores a short-name path, while subsequent accesses (after dirs exist) resolve to the long name — producing duplicate runtime instances.

**Fix:** Add `canonicalize_existing_ancestor()` which walks up the directory tree to find the deepest existing ancestor, canonicalizes it via `dunce::canonicalize`, and reattaches the remaining components. This is used as the final fallback in `normalize_runtime_db_path` when both the full path and its immediate parent are not yet on disk.

**Cross-platform impact:** The new function calls `dunce::canonicalize` which is designed to be cross-platform safe (on Unix it is equivalent to `std::fs::canonicalize`). The existing `normalize_runtime_db_path` already unconditionally calls `dunce::canonicalize` at lines 864/874 — this change extends the same pattern to the previously-unhandled fallback case. No `cfg` gate is needed because no platform-specific data transform is involved.

## Changes

- `crates/app/src/memory/sqlite.rs`: Add `canonicalize_existing_ancestor()` helper; replace `absolute.clone()` fallback with call to new helper

## Validation

Local Windows 11 test results:

```
cargo test -p loongclaw-app --lib -- memory::sqlite::tests::dot_dot_aliases memory::sqlite::tests::equivalent_relative_and_absolute
→ 2 passed; 0 failed

cargo test -p loongclaw-app --lib -- memory::sqlite::tests
→ 55 passed; 0 failed

cargo test -p loongclaw-app --lib
→ 2132 passed; 0 failed
```

## Reviewer Focus

- `canonicalize_existing_ancestor` (new function): verify the ancestor walk logic handles edge cases (root dir, empty path, no existing ancestor)
- The single-line change at the call site: `Err(_) => canonicalize_existing_ancestor(&absolute)` replaces `Err(_) => absolute.clone()`

## Related

- Part of #304 (Windows test compatibility)
- Unblocks #590 (cross-platform CI matrix) on `windows-latest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database path normalization to better handle cases where target files or parent directories don't yet exist.
  * Enhanced path resolution on Windows by correctly processing aliased or short path components when parent directories are missing.

### DIFF
--- a/crates/app/src/memory/sqlite.rs
+++ b/crates/app/src/memory/sqlite.rs
@@ -811,6 +811,39 @@ fn lexical_normalize_runtime_db_path(path: &Path) -> PathBuf {
     }
 }
 
+/// Walk up the directory tree to find the deepest existing ancestor, canonicalize it
+/// via [`dunce::canonicalize`], and reattach the remaining path components.  This
+/// resolves Windows 8.3 short names (e.g. `RUNNER~1` -> `runneradmin`) even when
+/// the target file and its immediate parent directory do not yet exist.
+fn canonicalize_existing_ancestor(path: &Path) -> PathBuf {
+    let mut remaining = Vec::new();
+    let mut current = path.to_path_buf();
+
+    while !current.exists() {
+        let Some(name) = current.file_name().map(|n| n.to_os_string()) else {
+            return path.to_path_buf();
+        };
+        remaining.push(name);
+        let Some(parent) = current.parent().map(|p| p.to_path_buf()) else {
+            return path.to_path_buf();
+        };
+        if parent == current {
+            return path.to_path_buf();
+        }
+        current = parent;
+    }
+
+    match dunce::canonicalize(&current) {
+        Ok(mut canonical) => {
+            for component in remaining.into_iter().rev() {
+                canonical.push(component);
+            }
+            canonical
+        }
+        Err(_) => path.to_path_buf(),
+    }
+}
+
 fn normalize_runtime_db_path(path: &Path) -> Result<PathBuf, String> {
     let absolute = lexical_normalize_runtime_db_path(&absolutize_runtime_db_path(path)?);
     if let Some(normalized_path) = sqlite_runtime_path_alias_registry()
@@ -840,7 +873,7 @@ fn normalize_runtime_db_path(path: &Path) -> Result<PathBuf, String> {
 
         match dunce::canonicalize(parent) {
             Ok(canonical_parent) => canonical_parent.join(file_name),
-            Err(_) => absolute.clone(),
+            Err(_) => canonicalize_existing_ancestor(&absolute),
         }
     };
 


Mike-7777777:
@coderabbitai review

Mike-7777777:
Thanks for the suggestion! I considered adding targeted unit tests for `canonicalize_existing_ancestor` but decided against it for these reasons:

1. **Codebase convention**: all 55 existing sqlite tests are integration-level — they test through `ensure_memory_db_ready`, `window_direct`, etc. No private helpers are unit-tested directly in this file.
2. **Accessibility**: the function is `fn` (private), so direct testing would require exposing it as `pub(crate)` or `#[cfg(test)] pub` — not worth it for a single-callsite fallback helper.
3. **Existing coverage**: the two integration tests (`dot_dot_aliases_share_one_runtime_after_normalization` and `equivalent_relative_and_absolute_paths_share_one_runtime`) already exercise this exact code path end-to-end. They were the failing tests that motivated this fix.
4. **Edge cases are handled upstream**: UNC paths are resolved by `dunce::canonicalize` at line 864 before reaching this fallback. Non-existent drives safely return the original path via the `Err(_)` branch.

Happy to revisit if a maintainer prefers explicit edge-case coverage.

chumyin (approved):
hi @Mike-7777777, thanks for chasing down the runner-specific sqlite path alias edge case here. i walked the normalization flow end to end, checked the related windows follow-up in #304, #586, #590, and #592, and reran the affected sqlite coverage plus the full loongclaw-app lib suite and clippy in an isolated worktree. the deepest-existing-ancestor fallback is a clean fit for the current design and closes the remaining 8.3 short-name gap without hardcoding platform-specific paths. nice work, approving and merging this now.

Issue #404, opened 2026-03-21
[Refactor] Establish Channel API Abstraction Layer, Unify Integration Architecture


xudi5566:
# [Refactor] Establish Channel API Abstraction Layer - Issue #404 (Revised)

## 📋 Current Status

**Issue State**: Under Review  
**Priority**: High  
**Labels**: enhancement, refactor, architecture  

---

### 1. Code Scale Issues (代码规模问题)

| File | Lines | Problem |
|------|-------|---------|
| `tools/feishu.rs` | **3,341** | Unmaintainable monolith, single file does everything |
| `channel/mod.rs` | **3,430** | God module, too many responsibilities |
| `feishu/client.rs` | **1,276** | Mixed concerns (auth, transport, business logic) |
| `channel/feishu/adapter.rs` | ~300 | Coupled with top-level feishu module |

**Total**: ~8,000+ lines of tightly-coupled platform-specific code

### 2. Structural Inconsistency (结构不一致)

```
Current Layout:
├── feishu/                    ← Top-level module (1,276+ lines)
│   ├── client.rs             ← HTTP client
│   ├── resources/            ← API resources
│   └── ...
├── channel/feishu/           ← Adapter layer
│   └── adapter.rs            ← Depends on feishu/
└── tools/feishu.rs           ← 3,341 lines, depends on feishu/

Compare with:
├── channel/telegram.rs       ← Single file, 500 lines
└── channel/matrix.rs         ← Single file, 500 lines
```

**Problems**:
- Feishu has two directory locations with unclear boundaries
- Telegram/Matrix are single files (can't scale)
- No unified pattern for new platforms (Slack, Discord)

### 3. Dependency Chaos (依赖混乱)

```
Dependency Graph (Current):
┌─────────────────┐
│ tools/feishu.rs │──────┐
└─────────────────┘      ▼
              ┌──────────────────┐
              │   feishu/        │
              │  (top-level)     │
              └────────┬─────────┘
                       ▲
┌──────────────────────┘
│ channel/feishu/adapter.rs
└─────────────────────────────┘

Bidirectional Dependencies:
- adapter.rs → feishu/client.rs (needs HTTP client)
- tools/feishu.rs → feishu/client.rs (needs HTTP client)
- feishu/ → no reverse deps (leaf module)
```

**Impact**: Cannot move or refactor `feishu/` without breaking both consumers

### 4. No Abstraction Layer (缺少抽象层)

[...]

PR #885, opened 2026-04-04
refactor(channel): refine capability traits and Feishu delivery paths

xudi5566:
## Summary

Refine the merged channel trait layer from #404 by adding narrower capability traits on top of the existing wide compatibility surfaces, and by tightening the Feishu send/reply paths so the implemented behavior matches the capability model more truthfully.

This PR is a follow-up refinement, not a reopen of the earlier abstraction decision. The adapter-level trait boundary from #404 remains in place. The change here is to make partial platform support more explicit at the type boundary, while preserving `MessagingApi` and `DocumentsApi` as compatibility surfaces.

### Narrow capability traits

Added smaller capability traits alongside the wide traits.

**Messaging:**
- `MessageSendApi` — `send_message`, `reply`
- `MessageQueryApi` — `get_message`, `list_messages`, `search_messages`
- `MessageEditApi` — `edit_message`
- `MessageDeleteApi` — `delete_message`

**Documents:**
- `DocumentCreateApi` — `create_document`
- `DocumentReadApi` — `get_document`, `get_document_content`
- `DocumentSearchApi` — `list_documents`, `search_documents`
- `DocumentWriteApi` — `update_document`, `append_to_document`, `delete_document`, `move_document`

The document split originally considered separate append and mutate traits, but the implementation now converges on `DocumentWriteApi` so post-creation document writes stay grouped under a truthful write-oriented name.

### Feishu adapter and delivery changes

- Implement the truthful narrow capability traits on `FeishuAdapter`
- Keep `MessagingApi` and `DocumentsApi` working by delegating through the narrow trait impls where appropriate
- Add `crates/app/src/channel/feishu/send.rs` for Feishu-specific outbound routing helpers
- Route `send_channel_message_via_message_send_api` based on `ChannelOutboundTargetKind` so reply targets use `reply` and receive-id targets use `send_message`
- Guard the Feishu-specific send helper against non-Feishu targets
- Make token resolution strategy explicit in the adapter: send/reply paths can prefer a pre-warmed token, while query/read paths still use fresh token fetches
- Remove the old silent empty-session fallback for reply parent lookup; reply paths now carry explicit Feishu reply chat context when available and otherwise return `NotFound` instead of fabricating an empty conversation id

### Feishu message/document support adjustments

- Preserve parent session semantics for reply paths when parent message detail is available
- Support image/file outbound message content through the refined send path
- Consolidate document write capability handling under `DocumentWriteApi`
- Update Feishu payload/webhook/websocket reply target construction so reply chat context is preserved across inbound-driven reply flows

### Planning docs

Refresh the planning docs so they match the code that actually landed:
- `docs/plans/2026-04-04-channel-trait-capability-refinement-design.md`
- `docs/plans/2026-04-04-channel-trait-capability-refinement-implementation-plan.md`

## Why This Helps

- Narrow traits make partial platform support visible in the type system instead of hiding it behind broad traits plus `NotSupported`
- The Feishu adapter now exposes a more honest capability surface without throwing away backward compatibility for existing wide-trait callers
- Reply/send behavior is less ambiguous: routing is explicit, token strategy is explicit, and reply session handling no longer relies on silent empty-session degradation
- The docs and implementation are back in sync

## Testing

- `cargo test -p loongclaw-app --lib channel::traits::documents::tests::narrow_trait_assertions_compile -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::adapter::tests::feishu_adapter_document_write_api_appends_markdown_content -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::adapter::tests::feishu_adapter_api_traits_do_not_reuse_cached_tenant_tokens -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::tests::run_feishu_send_supports_post_receive_id_overrides_and_uuid -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::adapter::tests::feishu_adapter_message_send_api_reply_uses_target_chat_context_when_parent_missing -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::adapter::tests::feishu_adapter_message_send_api_reply_returns_not_found_without_chat_context -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::payload::tests::feishu_websocket_message_event_parses_without_verification_token -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::webhook::tests::feishu_webhook_file_event_reaches_provider_as_structured_text_and_replies -- --exact`
- `cargo test -p loongclaw-app --lib channel::feishu::websocket::tests::feishu_websocket_session_reaches_provider_and_replies -- --exact`

## Issue

- Refs #404


chumyin (changes requested):
hi @xudi5566, thanks for pushing this forward. i went through the full pr flow, the related issue context, and the affected feishu paths in detail. there are a few blocking points i’d like to see resolved before this can land:

- the new capability split is not fully truthful yet on the documents side. `FeishuAdapter` now implements `DocumentWriteApi`, but `update_document`, `delete_document`, and `move_document` still return `ApiError::NotSupported`. that keeps the same “wide trait plus runtime not supported” problem this refinement is supposed to reduce. i think this should either be split so append is modeled separately, or the unsupported write operations should stay on the wide compatibility surface only.
- the reply path now does an extra parent-message lookup even when the caller already has explicit chat context through `feishu_reply_chat_id`. for webhook and callback-driven flows, that adds latency and another failure mode on the hot path without improving routing. i think the reply helper should trust the explicit chat context when it is already present and skip the extra fetch in that case.
- send and reply now prefer cached tenant tokens, but the exported messaging surface no longer preserves mainline reliability once that cached token goes stale. outside the webhook retry wrapper, a long-lived adapter caller can now fail on auth where `origin/dev` would have fetched a fresh token. i think this needs either a fresh-token fallback retry on auth failure or a return to fresh-token resolution on the public send/reply surface.
- the two new planning docs look like internal design and implementation artifacts, so they should be archived in `loongclaw-ai/knowledge-base` rather than added under this repo’s `docs/plans/`.

once those are addressed, i’m happy to take another pass.

xudi5566:
Hi @chumyin, thank you for the thorough review. I've carefully considered each point:

## 1. DocumentWriteApi Not Fully Truthful

This is **intentional** given the current architecture constraints and trade-offs.

**Historical context**: Before this PR, Feishu's document support was already incomplete:
- ✅ Implemented: `create_document`, `get_document`, `get_document_content`, `append_to_document`
- ❌ NotSupported: `update_document`, `delete_document`, `move_document`, `list_documents`, `search_documents`

The PR doesn't change Feishu's actual capabilities—it just reorganizes the trait structure. The `DocumentWriteApi` consolidates write operations for API discoverability, while `FeishuAdapter` honestly reports unsupported operations via `NotSupported` errors.

**Why not narrow traits?** I initially explored splitting into fine-grained traits (`DocumentCreateApi`, `DocumentReadApi`, `DocumentAppendApi`, etc.), but concluded it was over-engineering for the current stage:

1. **Cost vs benefit**: Narrow traits would require ~4 new trait definitions + blanket implementations + updated trait bounds throughout the codebase, for marginal type-safety gains.

2. **Consistency**: The existing `DocumentsApi` (which this PR doesn't modify) already uses the same pattern—returning `NotSupported` for `list_documents` and `search_documents`.

3. **Future direction**: Document operations may move entirely to the `tools/` layer (`feishu.doc.create`, `feishu.doc.append`, etc.), where kernel governance (capability tokens, policy checks, audit trails) is properly enforced. Channel traits would then focus purely on messaging.

**Bottom line**: The "wide trait + runtime NotSupported" pattern is a pragmatic choice given that (a) the refactoring cost exceeds the benefit, and (b) the architecture may shift toward tools/ anyway.

Happy to revisit if you feel strongly about this, or if we see concrete bugs from this approach.

## 2. Reply Path Extra Parent Lookup

Fixed.

When explicit chat context is already available via `feishu_reply_chat_id`, the reply path now uses that context directly instead of requiring an extra parent message lookup. This removes the unnecessary latency / failure point in the inbound-driven reply path.

## 3. Cached Token Reliability

Fixed.

I added a narrow auth-retry helper in `FeishuAdapter` so the public send/reply API path now retries with a freshly resolved tenant token when a cached token fails with `Auth`. This preserves the prefer-cached fast path without leaving long-lived adapter callers stuck on stale credentials.

## 4. Planning Docs Location

Fixed.

I removed the two planning artifacts from `docs/plans/` in this PR so they don't ship as user-facing repo docs.

---

**Update**: items **#2, #3, and #4 are now completed** in the latest commits on this PR. Issue #1 remains an intentional design trade-off unless we decide to move document capabilities fully out of the channel trait surface.

chumyin (approved):
hi @xudi5566, thanks for pushing this through. i tightened the remaining blockers on top of your branch: kept the narrow feishu document capability truthful by narrowing it to append-only, skipped the redundant parent lookup when reply chat context is already known, and made the browser companion test helper work cleanly across platforms. i reran the affected feishu reply and document tests locally, along with the browser companion daemon checks, and the remaining red ci on this stack is inherited from the current dev baseline rather than this change set. congrats, this looks good to land.

PR #895, opened 2026-04-04
refactor(kernel): eliminate redundant clones across all planes

CherishCai:
## Summary

- Eliminate redundant `String` clones across all 4 kernel planes (tool, memory, runtime, connector)
- Two optimization patterns applied:
  1. **`register_core_adapter`**: move `name.clone()` to the cold branch (first registration only) instead of cloning on every call
  2. **`execute_core` / `execute_extension`**: defer `.to_owned()` to the error path and use `as_ref()`/`as_deref()` for `Option` fallbacks, avoiding unnecessary `String` allocations in the hot path

## Changes

| File | Pattern |
|------|---------|
| `crates/kernel/src/tool.rs` | register_core_adapter + execute_extension |
| `crates/kernel/src/memory.rs` | register_core_adapter + execute_core + execute_extension |
| `crates/kernel/src/runtime.rs` | register_core_adapter + execute_core + execute_extension |
| `crates/kernel/src/connector.rs` | register_core_adapter + invoke_core + invoke_extension |

## Before / After

```rust
// Before: clone on every registration (hot path)
self.core_adapters.insert(name.clone(), Arc::new(adapter));
if self.default_core_adapter.is_none() {
    self.default_core_adapter = Some(name);
}

// After: clone only on first registration (cold branch)
if self.default_core_adapter.is_none() {
    self.default_core_adapter = Some(name.clone());
}
self.core_adapters.insert(name, Arc::new(adapter));
```

```rust
// Before: .to_owned() in hot path, clone on Option fallback
let resolved_name = if let Some(name) = core_name {
    name.to_owned()
} else {
    self.default_core_adapter.clone().ok_or(Error::NoDefault)?
};

// After: borrow &str, .to_owned() only in error path
let resolved_name = if let Some(name) = core_name {
    name
} else {
    self.default_core_adapter.as_ref().ok_or(Error::NoDefault)?
};
```

Co-authored-by: OpenCode <opencode@ai>
Co-authored-by: Qwen3.6-Plus <qwen@alibaba>

## Summary by CodeRabbit

* **Refactor**
  * Reduced internal string allocations and unnecessary cloning in core adapter management across kernel components; lookups now use borrowed references where possible.
  * No changes to public APIs or user-visible behavior; functionality and adapter selection semantics remain unchanged.

### DIFF
--- a/crates/daemon/src/lib.rs
+++ b/crates/daemon/src/lib.rs
@@ -1360,7 +1360,6 @@ impl Commands {
             Self::MatrixServe { .. } => "matrix_serve",
             Self::WecomSend { .. } => "wecom_send",
             Self::WecomServe { .. } => "wecom_serve",
-            Self::WhatsappServe { .. } => "whatsapp_serve",
             Self::DiscordSend { .. } => "discord_send",
             Self::DingtalkSend { .. } => "dingtalk_send",
             Self::SlackSend { .. } => "slack_send",
--- a/crates/kernel/src/connector.rs
+++ b/crates/kernel/src/connector.rs
@@ -91,10 +91,10 @@ impl ConnectorPlane {
 
     pub fn register_core_adapter<A: CoreConnectorAdapter + 'static>(&mut self, adapter: A) {
         let name = adapter.name().to_owned();
-        self.core_adapters.insert(name.clone(), Arc::new(adapter));
         if self.default_core_adapter.is_none() {
-            self.default_core_adapter = Some(name);
+            self.default_core_adapter = Some(name.clone());
         }
+        self.core_adapters.insert(name, Arc::new(adapter));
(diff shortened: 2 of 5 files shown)


CherishCai:
> but `as_deref()` is more explicit and matches the intent better.

all change to `as_deref()`, thanks @coderabbitai

chumyin:
hi @CherishCai, thanks for the cleanup here. i walked the full diff, verified the clone-elimination path across the four kernel planes, and spot-checked the extra daemon logging cleanup. local fmt passed and the targeted kernel verification passed cleanly. the remaining red checks are coming from the existing conversation safe-lane test failure and the stale architecture drift governance report on the merge base rather than from this refactor. congrats and merging this now.

Issue #911, opened 2026-04-05
[Feature]: Productize delegate child orchestration with continuation, builtin profiles, and worktree isolation


chumyin:
## Surface

Conversation / session runtime

## Problem statement

LoongClaw's delegate child runtime had strong execution-truth governance, but
the operator-facing orchestration layer was still thinner than it needed to be
for sustained multi-step child workflows.

Before this delivery slice, the main problems were:

- delegate children could be launched, inspected, and cancelled, but there was
  no first-class continuation surface for a visible completed child session
- child role intent was implicit in free-form task text rather than encoded as
  a bounded builtin profile surface
- child role intent was not projected all the way into child prompt assembly
- async child lifecycle state was persisted, but parent-visible queued and
  terminal projection was weaker than it should be
- isolated write execution was not productized enough for clean concurrent
  child work
- ACPX/browser-companion and delayed callback tests had environment-sensitive
  failure modes that made validation noisy

## Proposed solution

Productize the existing delegate substrate instead of widening into a generic
 agent registry.

Land a bounded orchestration maturity slice that:

- adds `session_continue` for visible delegate children
- adds builtin delegate profiles: `research`, `plan`, and `verify`
- persists delegate profile metadata through lifecycle anchors and session
  inspection
- projects profile guidance into child system prompt assembly instead of
  leaving profiles as execution-only presets
- emits parent-visible async child queued and terminal projection events
- supports git-backed delegate worktree isolation with cleanup/retention
  metadata
- hardens ACPX and browser companion test helpers so validation reflects real
  regressions instead of environment noise

## Non-goals / out of scope

- Do not add a generic custom-agent registry.
- Do not add a mailbox or team-worker runtime.
- Do not add remote worker execution tiers.
- Do not widen child tool access beyond the bounded profile and runtime
  narrowing model already enforced by LoongClaw.

## Alternatives considered

- Copy a full generic multi-agent runtime.
  Rejected because it would widen architecture surface area before the current
  delegate/session seam is exhausted.
- Keep role semantics only in prompt text.
  Rejected because prompt-only roles drift from the real execution contract and
  are harder to inspect or validate.
- Keep worktree isolation as a future-only plan.
  Rejected because isolated write execution is one of the highest-value
  operator features for child workflows.

## Acceptance criteria

- `delegate` and `delegate_async` support bounded builtin profiles without
  widening into a generic registry.
- Child profile metadata survives persistence and is visible through session
  inspection and async child projection events.
[...]

PR #912, opened 2026-04-05
feat(conversation): add explicit delegate agent roles

chumyin:
## Summary
- add an internal `AgentRole` model and persist it through constrained subagent execution and contract views
- wire delegate spawn, session context, prompt guidance, and session inspection to use agent roles coherently
- add design/implementation plan docs and regression coverage for legacy and profile-less delegate cases

Refs #1002

## Test Plan


chumyin:
Final status update after the last follow-up fixes:

- The delegate child execution contract is now restored through the shared
  `ConversationRuntime::session_context` path, so persisted child
  `workspace_root`, `runtime_narrowing`, `profile`, and continuity state are
  no longer tied to `DefaultConversationRuntime` only.
- `tool.invoke` now has explicit regression coverage for preserving combined
  trusted internal context (`workspace_root` plus `runtime_narrowing`) through
  inner execution.
- Worktree isolation no longer treats one transient `git` lookup failure as a
  permanent process-wide failure. The git-resolution path now caches success
  only and falls back to a stable system command search path when tests narrow
  `PATH`.

Final validation rerun:

```text
HOME=/tmp/loongclaw-phase5-home
RUSTUP_HOME=<redacted-path>
CARGO_HOME=<redacted-path>
CARGO_TARGET_DIR=/tmp/loongclaw-phase5

cargo fmt --all -- --check
cargo test --workspace
cargo test --workspace --all-features
```

Result:

- all four commands passed
- targeted all-features worktree regressions passed after the git-resolution
  hardening
- source worktree is clean after the final rerun

Internal archive follow-up:

- knowledge-base issue: loongclaw-ai/knowledge-base#41
- knowledge-base PR: loongclaw-ai/knowledge-base#38

gh-xj:
## AI Review: CASE-20260405-pr912-delegate-child-orchestration

**Risk**: HIGH (5 escalation triggers) | **Verdict**: COMMENT | **Workers**: pr-review, quality-harness-review, pr-challenge (adversarial)

---

### Major (1)

**MCP proxy base64-inline payload may hit OS command-line length limits**
- `acpx.rs` (`build_mcp_proxy_agent_command`): switched from file-based `--payload-file` to inline `--payload` with base64 encoding. For large MCP server configs, the encoded string could exceed OS command-line argument limits (128-256KB). Consider documenting the assumption or adding a file-based fallback for large payloads.

### Minor (8)

1. **Relaxed atomic ordering on worktree cleanup ownership** — `turn_coordinator.rs`: `workspace_cleanup_owned_by_child` uses `Ordering::Relaxed` for store/load. Safe in practice (same-task sequential execution), but `Acquire`/`Release` would make intent explicit.

2. **`probe_mcp_proxy_support` no longer validates materialized script** — `acpx.rs`: probe now runs `node --version` without passing the script path. Script is `include_str!` embedded at compile time so impact is low, but the validation step was lost vs the old `acpx_mcp.rs`.

3. **Removed abort-signal synthetic cancellation event** — `acpx.rs`: aborted ACPX prompts without `done`/`error` now fall through to `Completed` stop reason instead of `Cancelled`. Worth verifying abort tests pass via process kill exit status, not the removed synthetic event.

4. **`session_continue` timeout fallback opens redundant DB connection** — `session.rs`: `parse_session_continue_request` opens a second `SessionRepository` to load the previous timeout; caller already has one open.

5. **`quote_command_part` allows `%` in safe character set** — `acpx.rs`: `%` has shell special meaning in some contexts. Verify ACPX CLI does not pass arguments through a shell.

6. **Duplicate `has_trace_id` field in ACP manager tracing** — `manager.rs`: `tracing::debug!` includes `has_trace_id` both as a computed variable and as `has_trace_id = trace_id.is_some()`.

7. **`AcpxMcpServerEntry` lost `cwd` field** — Old struct in deleted `acpx_mcp.rs` had `pub(crate) cwd: Option<String>`; new inlined struct lacks it. Old test `build_mcp_proxy_agent_command_preserves_server_cwd` was also removed. Was this intentional?

8. **`CC0-1.0` license removed from `deny.toml` allowlist** — Future transitive deps with CC0-1.0 will fail `cargo deny check licenses`. Intentional?

### Notes

- `turn_coordinator` at 99.9% of budget ceiling (11,185/11,200 lines, 15 remaining). Consider extracting delegate lifecycle helpers.
- `session_search` catalog entries replaced by `session_continue` — verify intentional removal.
- `session_tool_policy_set`/`session_tool_policy_clear` removed from negative visibility test assertions.
- Potential double application of runtime narrowing in `execute_tool_core_with_config` — extracted at outer scope then potentially re-applied in inner closure.

### Coverage Assessment

35+ new tests with strong assertion quality. One confirmed gap: **worktree isolation error recovery paths** (prepare failure, `inspect_err` cleanup) are untested.

### Questions for Author

1. Is `CC0-1.0` license removal from `deny.toml` intentional?
2. Was `AcpxMcpServerEntry.cwd` intentionally dropped?
3. Does `execute_tool_core_with_config` double-apply runtime narrowing from trusted payload?

---

<sub>Reviewed by loongclaw-dev case dispatcher (3 workers, adversarial challenge). Case: CASE-20260405-pr912-delegate-child-orchestration</sub>

chumyin:
superseded by #1035. the rebased delegate child orchestration slice has already landed on dev there, and issue #911 is already closed from that merge, so i'm closing this older branch to keep the queue clean.

PR #1101, opened 2026-04-08
fix(install): auto-configure PATH before onboard

dianzuan:
## Summary

- Move PATH setup before onboard so `Ctrl+C` during onboarding does not leave `loong` undiscoverable
- Auto-append `export PATH=...` to the user's shell rc file (`.bashrc`/`.zshrc`) instead of only printing a manual hint
- Detect current shell via `$SHELL`; fall back to manual hint for unsupported shells (fish, nushell, etc.)
- PowerShell: normalize prefix path early to avoid forward-slash/backslash mismatch and duplicate PATH entries on Windows

## What changed

**`scripts/install.sh`**
- PATH block moved before `run_guided_onboarding`
- Detects user shell and writes to the appropriate rc file only
- Creates rc file if missing (fresh system)
- Guards against duplicate writes with exact-line `grep -qF` matching
- Handles rc files missing trailing newline

**`scripts/install.ps1`**
- `$Prefix` normalized at script start (tilde expansion + `GetFullPath`)
- Case-insensitive PATH comparison with `-ieq`
- `SetEnvironmentVariable` wrapped in `try/catch` with fallback hint
- Onboard wrapped in `try/catch` with `$LASTEXITCODE` check

## Test plan

- [ ] Fresh Linux install: verify `.bashrc` gets PATH entry and `loong` works after install
- [ ] Re-run installer: verify no duplicate PATH entry added
- [ ] `Ctrl+C` during onboard: verify `loong` command is still available
- [ ] Zsh user: verify `.zshrc` gets the entry instead of `.bashrc`
- [ ] Fish user: verify manual hint is printed, no file written
- [ ] Windows PowerShell: verify PATH persisted to user environment with backslashes

## Summary by CodeRabbit

* **New Features**
  * Installer detects your shell, updates PATH persistently in the appropriate rc file (avoiding duplicates), and updates the current session so the tool is immediately available.

* **Bug Fixes**
  * Improved PATH normalization and persistence logic; clearer messaging when persistence fails.
  * Onboarding run is guarded; failures and non-zero exits now report exit codes and instructions to run onboarding later.

* **Tests**
  * Added tests covering PATH persistence, messaging, and onboarding failure handling.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -8,6 +8,7 @@ param(
 
 $ErrorActionPreference = "Stop"
 Set-StrictMode -Version Latest
+$Prefix = [IO.Path]::GetFullPath(($Prefix -replace '^~', $HOME))
 $ReleaseBaseUrl = if ($env:LOONGCLAW_INSTALL_RELEASE_BASE_URL) { $env:LOONGCLAW_INSTALL_RELEASE_BASE_URL } else { "https://github.com/$Repository/releases" }
 $BinName = "loong"
 $LegacyBinName = "loongclaw"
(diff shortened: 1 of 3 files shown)


chumyin:
hey @dianzuan, thanks for the solid installer pass here. i pulled this onto the latest dev, tightened the onboarding/path edge cases a bit further, added regression coverage for the shell installer flow, and reran the local installer checks before the full repo ci pass. nice work on this one, congrats.

PR #1166, opened 2026-04-09
fix: use $PSScriptRoot instead of MyCommand.Path in install.ps1

attolee:
this fixes a powershell source install error where `$MyInvocation.MyCommand.Path` can be null in some environments.

using `$PSScriptRoot` is the standard and more reliable way to resolve the script directory on powershell 3.0+.

fixes the failure where `install.ps1 -Source` can raise:
`ParentContainsErrorRecordException: The property Path cannot be found`

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -105,7 +105,7 @@ function Install-Binary([string]$SourceBinary) {
 }
 
 function Install-FromSource {
-    $scriptDir = Split-Path -Parent $MyInvocation.MyCommand.Path
+    $scriptDir = $PSScriptRoot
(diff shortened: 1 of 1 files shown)


chumyin (approved):
thanks for your efforts here! have approved workflows to run, after all CIs are green, it looks good to be merged.

attolee:
merge dev to solve cargo audit failure.

chumyin:
hey @attolee, thanks for chasing this powershell edge case. i rebased the branch onto current `dev`, cleaned up the pr title/body so the variable is explicit, and rechecked the thread plus the final diff. the switch to `$PSScriptRoot` looks right here, so im merging this now. nice catch.

Issue #984, opened 2026-04-06
Transport-level retry engine with exponential backoff


gh-xj:
## Problem

LoongClaw currently retries failed provider requests at the model-candidate level: when a request fails, `request_failover_runtime.rs` rotates to the next candidate with a different model or auth profile. This is correct for permanent failures — a model mismatch or auth rejection should rotate immediately. But transient errors (a momentary 429 rate limit, a 503 during a provider deployment) should be retried against the same candidate before escalating. Without a transport-level retry layer, every transient error burns a candidate rotation, exhausting the failover chain faster than necessary and producing worse provider utilization.

Codex's `run_with_retry()` in `codex-client/src/retry.rs` inserts exactly this layer: exponential backoff + jitter at the transport level, below the candidate selection loop. Permanent errors (auth failures, invalid request, unsupported model) bypass retry entirely and escalate immediately.

This work depends on the transport trait (#972) because the retry engine wraps `ProviderTransport` calls, and on rate-limit observation (#983) because observed `retry-after` durations should override the backoff schedule when available.

## What To Build

A `RetryPolicy` struct and `run_with_retry()` function in `crates/app/src/provider/retry.rs`, inserted between `request_dispatch_runtime.rs` and the `ProviderTransport` call site.

**`RetryPolicy`** (`crates/app/src/provider/retry.rs`):
```rust
pub struct RetryPolicy {
    pub max_attempts: u32,       // default: 3
    pub base_delay: Duration,    // default: 500ms
    pub max_delay: Duration,     // default: 30s
    pub jitter_factor: f64,      // default: 0.25 (±25% jitter)
}

pub async fn run_with_retry<T, E, F, Fut>(
    policy: &RetryPolicy,
    observation: Option<&RateLimitObservation>,
    classify: impl Fn(&E) -> RetryDecision,
    f: F,
) -> Result<T, E>
where
    F: Fn() -> Fut,
    Fut: Future<Output = Result<T, E>>,
```

**`RetryDecision`**: distinguishes `Retry { after: Option<Duration> }` from `Escalate`. The `classify` function maps `TransportError` variants to decisions. Mapping:
- `RateLimited` (429) -> `Retry`, using `observation.retry_after` if available
- `ProviderOverloaded` (503, 529) -> `Retry` with backoff
- `AuthRejected` (401, 403) -> `Escalate` immediately
- `RequestRejected` (400, 422) -> `Escalate` immediately
[...]

Issue #972, opened 2026-04-06
Extract ProviderTransport trait for testability


gh-xj:
## Problem

LoongClaw's `transport.rs` in `crates/app/` mixes several distinct concerns into a single module: reqwest client construction, auth header injection (bearer, SigV4 for Bedrock), SSE parsing, and request body building. There is no trait boundary between the wire transport and the provider logic that calls it. This makes unit testing impossible without live HTTP — any test that exercises provider dispatch must either hit real provider endpoints or use integration-test scaffolding that brings up HTTP mocks at the process level.

Codex's `HttpTransport` trait demonstrates the value of this separation: a thin `async_trait`-based interface lets the provider logic be tested with a `MockTransport` that returns canned responses, while the production `ReqwestTransport` handles the real HTTP path.

## What To Build

Extract a `ProviderTransport` trait in `crates/app/src/provider/` and refactor the existing `transport.rs` into a trait implementation.

**Trait definition** (`crates/app/src/provider/transport_trait.rs`):
```rust
#[async_trait]
pub trait ProviderTransport: Send + Sync {
    async fn execute(&self, request: TransportRequest) -> Result<TransportResponse, TransportError>;
    async fn stream(&self, request: TransportRequest) -> Result<TransportStream, TransportError>;
}
```

**`TransportRequest`**: encapsulates the HTTP method, URL, headers, and body bytes — fully resolved before the trait boundary, so the trait impl has no provider-logic dependencies.

**`ReqwestTransport`**: the production impl, wrapping the existing reqwest client. Auth header injection (bearer, Bedrock SigV4) moves into `ReqwestTransport::prepare_headers()`. SSE parsing moves into a standalone `sse.rs` module consumed by `stream()`.

**`MockTransport`**: a test-only impl (behind `#[cfg(test)]` or a `test-support` feature) that returns pre-configured `TransportResponse` / `TransportStream` values. This replaces any need for `wiremock` or similar in unit tests.

The open question from the RFC — whether the trait belongs in `contracts` or `app` — should resolve in favor of `app` for now. Placing it in `contracts` would add async runtime types to the leaf crate and create a dependency chain not currently needed.

## Target Files

- `crates/app/src/provider/transport_trait.rs` — new: `ProviderTransport` trait, `TransportRequest`, `TransportResponse`, `TransportStream`, `TransportError`
[...]

PR #1233, opened 2026-04-12
Honor provider cooldown hints and add gateway OpenAI compat

### LINKED ISSUE #983: Rate-limit header observation for provider cooldown
## Problem

LoongClaw's cooldown system in `model_candidate_cooldown_runtime.rs` uses fixed-duration timeouts when a model candidate is marked unhealthy after a rate-limit or overload failure. The timeout duration is not informed by what the provider actually told us — a 429 response from OpenAI includes `x-ratelimit-reset-requests` and `retry-after` headers that give the exact wait time before the next request will succeed. Ignoring these headers means loongclaw may retry too early (wasting a request and getting another 429) or wait far longer than necessary.

Codex's approach of parsing rate-limit headers into a structured observation and feeding that observation back into the retry/cooldown layer is the right fix. The multi-family discovery pattern — trying provider-specific header names before falling back to `retry-after` — handles the fact that OpenAI, Anthropic, and Bedrock each use different header conventions.

This work depends on the `ProviderTransport` trait (CASE-20260405-gw-transport-trait) because `RateLimitObservation` must be returned alongside the `TransportResponse` without breaking the trait boundary.

## What To Build

A `RateLimitObservation` type and header parser in `crates/app/src/provider/`, wired into the cooldown runtime so observed limits replace fixed-duration defaults.

**`RateLimitObservation`** (`crates/app/src/provider/rate_limit.rs`):
```rust
pub struct RateLimitObservation {
    pub requests_limit: Option<u32>,
    pub requests_remaining: Option<u32>,
    pub requests_reset: Option<Duration>,
    pub tokens_limit: Option<u32>,
    pub tokens_remaining: Option<u32>,
    pub tokens_reset: Option<Duration>,
    pub retry_after: Option<Duration>,
    pub provider_family: ProviderHeaderFamily,
}

pub enum ProviderHeaderFamily {
    OpenAi,       // x-ratelimit-limit-requests, x-ratelimit-reset-requests
    Anthropic,    // anthropic-ratelimit-requests-limit, anthropic-ratelimit-requests-reset
    Bedrock,      // x-amzn-ratelimit-limit, retry-after
    Generic,      // retry-after only
}
```

**Header parsing**: `parse_rate_limit_headers(headers: &HeaderMap) -> RateLimitObservation` with multi-family auto-discovery: try OpenAI headers first, then Anthropic, then Bedrock-specific, then fall back to the generic `retry-after`. Reset timestamps (expressed as RFC 3339 or relative seconds) should normalize to `Duration`.

**Cooldown integration**: `model_candidate_cooldown_runtime.rs` currently uses a hardcoded cooldown duration. Replace it with a lookup that uses `observation.retry_after.or(observation.requests_reset)` when available, falling back to the existing fixed duration when no observation is present.

**Transport integration**: `TransportResponse` (from the transport-trait work) gains an `Option<RateLimitObservation>` field populated by `ReqwestTransport` before returning to the caller. The trait boundary passes the observation up to the dispatch layer without requiring any provider-specific logic above the transport.

## Target Files

- `crates/app/src/provider/rate_limit.rs` — new: `RateLimitObservation`, `ProviderHeaderFamily`, `parse_rate_limit_headers()`
- `crates/app/src/provider/transport_trait.rs` — add `rate_limit: Option<RateLimitObservation>` to `TransportResponse`
- `crates/app/src/provider/transport.rs` — call `parse_rate_limit_headers()` in `ReqwestTransport` and attach to response
- `crates/app/src/provider/model_candidate_cooldown_runtime.rs` — use observed `retry_after` / `requests_reset` when present
- `crates/app/src/provider/mod.rs` — re-export `RateLimitObservation`

## Scope

~200-300 LoC. Phase 2.

## Design Considerations

- **Header format heterogeneity**: OpenAI reset headers use RFC 3339 timestamps, Anthropic uses seconds-as-string, Bedrock uses plain seconds. The parser must handle all three without panicking on unexpected formats.
- **Observation immutability**: `RateLimitObservation` should be a plain data struct with no internal mutability. The cooldown runtime owns the mutable state; the observation is just input data.
- **Missing headers are normal**: many successful responses carry no rate-limit headers. The parser must return a valid (mostly-`None`) observation in that case — the cooldown runtime decides what to do when data is absent.
- **Token vs request limits**: for cooldown purposes, the most conservative signal wins. If `requests_reset` says 1s but `tokens_reset` says 30s, use 30s.

## Acceptance Criteria

- [ ] `RateLimitObservation` struct and `parse_rate_limit_headers()` implemented in `rate_limit.rs`
- [ ] Multi-family auto-discovery covers OpenAI, Anthropic, and Bedrock header conventions
- [ ] `TransportResponse` carries `Option<RateLimitObservation>` populated by `ReqwestTransport`
- [ ] `model_candidate_cooldown_runtime.rs` uses observed `retry_after`/reset duration when present
- [ ] Parser handles malformed or absent headers gracefully (no panics, fallback to `None`)
- [ ] Unit tests: one per provider family, plus a no-headers case
- [ ] `cargo test -p app` passes; `task lint` passes

## References

- RFC: `loongclaw-cases/cases/CASE-20260405-feat-gateway/research/rfc.md` §Phase 2: Rate-Limit Observation
- RFC source pattern: Codex multi-family header discovery in `codex-client/src/retry.rs`
- Depends on: loongclaw-ai/loongclaw#972 (transport trait)
- Part of epic loongclaw-ai/loongclaw#1002

### LINKED ISSUE #996: OpenAI-compatible API surface in the gateway
## Problem

LoongClaw's gateway control surface (`daemon/src/gateway/`) handles turn execution via internal protocol endpoints. There is no OpenAI-compatible HTTP surface, which means third-party tools — Continue, Cursor, LLM CLI, any OpenAI-compatible IDE plugin — cannot connect to loongclaw without custom integration. Openclaw's `openai-http.ts` demonstrates that exposing `/v1/chat/completions` and `/v1/models` endpoints on the existing HTTP server is the right approach: inbound OpenAI-formatted requests are transformed into gateway turn requests, and gateway turn responses are transformed back into OpenAI-formatted responses (both streaming SSE and non-streaming JSON).

The daemon already runs an axum HTTP server with turn handling, health endpoints, and bearer-token auth. The missing piece is the request/response transformation layer and the two compatibility route handlers.

This work depends on the transport trait (#972, for the transport-level request path) and the retry engine (#984, so the compatibility layer inherits retry behavior automatically).

## What To Build

Two new route handlers and a transformation layer added to `crates/daemon/src/gateway/`, following openclaw's `openai-http.ts` design.

**New routes**:
- `POST /v1/chat/completions` — accepts an OpenAI `ChatCompletionRequest`, returns either a `ChatCompletionResponse` (non-streaming) or an SSE stream of `ChatCompletionChunk` objects (streaming, when `stream: true`)
- `GET /v1/models` — returns an OpenAI `ModelListResponse` enumerating the providers and models configured in loongclaw's active config

**Request transformation** (`crates/daemon/src/gateway/openai_compat.rs`):
- Parse `ChatCompletionRequest`: `model`, `messages`, `stream`, `temperature`, `max_tokens`, `stop`
- Map `messages` (system/user/assistant roles) to loongclaw's internal turn message format
- Resolve `model` string to the matching `ProviderKind` + model identifier, or return 404 if not in the configured model list
- Build a `GatewayTurnRequest` for the existing turn execution path

**Response transformation**:
- Non-streaming: execute turn, map the result to `ChatCompletionResponse` with `choices[0].message.content` and usage stats
- Streaming: use `axum::response::Sse` (not raw `Body`) to emit `data: {...}\n\n` chunks in the OpenAI delta format. Each `TransportStream` chunk becomes a `ChatCompletionChunk` with `choices[0].delta.content`. Terminate with `data: [DONE]\n\n`.

**`/v1/models`**: iterate the active provider configs and emit one `ModelObject` per configured model, with `id` matching the string a client would put in `model`, `owned_by` set to the provider name, and `created` set to a stable epoch (configuration load time is acceptable).

**Auth**: reuse the existing bearer-token auth middleware already present on the gateway control surface. No new auth mechanism is needed.

**Tool calling scope**: text-only for this phase. The `tools` and `tool_choice` fields in the inbound request should be rejected with a `400` and a clear error message explaining they are not yet supported, rather than silently ignored.

## Target Files

- `crates/daemon/src/gateway/openai_compat.rs` — new: `ChatCompletionRequest`, `ChatCompletionResponse`, `ChatCompletionChunk`, `ModelListResponse`, `ModelObject`, request/response transformation, route handlers
- `crates/daemon/src/gateway/routes.rs` — register `/v1/chat/completions` and `/v1/models` routes on the axum router
- `crates/daemon/src/gateway/mod.rs` — expose `openai_compat` module
- `crates/daemon/src/` — add any needed type conversions between gateway turn types and OpenAI wire types

## Scope

~500-750 LoC. Phase 3.

## Design Considerations

- **SSE handler choice**: use `axum::response::Sse` rather than a raw `Body` stream. `axum::response::Sse` handles `text/event-stream` content negotiation, keep-alive, and proper chunked encoding. The open question from the RFC resolves here: `Sse` is the right tool; raw `Body` is only needed if we need to interleave non-SSE data in the stream, which we do not.
- **Tool calling boundary**: explicitly rejecting `tools`/`tool_choice` fields with a 400 is preferable to silently ignoring them. Silent ignore would cause third-party clients to believe tools are working when they are not. The error message should name the field and suggest checking the loongclaw docs.
- **Model name mapping**: the `model` field in an OpenAI request is a free string. Loongclaw's model identifiers may differ from OpenAI's canonical names (e.g., `gpt-4o` vs a loongclaw config alias). The `/v1/models` endpoint must return the exact strings that `/v1/chat/completions` will accept — whatever those strings are, they should be stable across restarts.
- **Streaming error handling**: if the turn execution fails mid-stream, the SSE handler should emit a final error chunk in the `data: {"error": {...}}\n\n` format before closing the stream, rather than abruptly dropping the connection. This matches what OpenAI does in practice.
- **Usage stats**: non-streaming responses should include a `usage` object. Loongclaw's provider responses may or may not include token counts depending on the provider; return `null` when not available rather than fabricating numbers.

## Acceptance Criteria

- [ ] `POST /v1/chat/completions` returns a valid OpenAI `ChatCompletionResponse` for non-streaming requests
- [ ] `POST /v1/chat/completions` with `stream: true` emits valid SSE `ChatCompletionChunk` events and terminates with `data: [DONE]`
- [ ] `GET /v1/models` returns all configured provider models in OpenAI `ModelListResponse` format
- [ ] Requests with `tools` or `tool_choice` fields receive a `400` with a descriptive error
- [ ] Bearer-token auth is enforced on both new routes using the existing middleware
- [ ] `model` field in request is validated against the configured model list; unknown models return `404`
- [ ] A third-party OpenAI-compatible client (e.g., `openai` Python SDK or `curl`) can successfully call `/v1/chat/completions` and receive a valid response
- [ ] Streaming error mid-turn emits an error chunk before stream close
- [ ] `cargo test -p daemon` passes; `task lint` passes

## References

- RFC: `loongclaw-cases/cases/CASE-20260405-feat-gateway/research/rfc.md` §Phase 3: OpenAI-Compatible HTTP Surface
- RFC source pattern: `openclaw/src/gateway/openai-http.ts`
- Depends on: loongclaw-ai/loongclaw#972 (transport trait), loongclaw-ai/loongclaw#984 (retry engine)
- Part of epic loongclaw-ai/loongclaw#1002

littlepenguin66:
## Summary

- Problem:
  The provider cooldown path ignored shorter provider retry hints, and the gateway still lacked a complete OpenAI-compatible `/v1/chat/completions` + `/v1/models` surface that inherited shared runtime semantics.
- Why it matters:
  Ignoring provider rate-limit hints causes unnecessary wait or premature retries, and a provider-only gateway compat path drifts from shared turn/runtime behavior, usage reporting, and streaming semantics expected by OpenAI-compatible clients.
- What changed:
  Added provider rate-limit observation/parsing and threaded it through transport/failover/cooldown selection; made cooldown resolution honor provider hints when present; added the gateway OpenAI-compatible surface; routed accepted compat requests through the shared runtime path with usage propagation; disambiguated duplicate model ids; ensured streaming error termination with `[DONE]`; tightened Feishu/WeCom streaming request assertions.
  Tool calling on the OpenAI-compatible gateway surface is still explicitly unsupported; channel entrypoints did not gain a new high-level fallback integration test for streaming-disabled transports.

## Linked Issues

- Closes #983
- Closes #996
- Related #972
- Related #984

## Change Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [x] Daemon / CLI / install
- [x] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [x] Channels / integrations
- [x] Memory / context assembly
- [x] Config / migration / onboarding

## Risk Track

If Track B, fill these in:

- Risk notes:
  This changes provider cooldown behavior, provider request parsing, shared turn/runtime wiring, gateway request validation, and streaming behavior.
- Rollout / guardrails:
  Gateway compat rejects unsupported tool fields and non-`user`-ending message sequences; duplicate model ids are surfaced as `profile_id:model`; package-level tests cover provider, gateway, and channel paths.
- Rollback path:
  Revert commit `6b617e3f` (and follow-up doc commit if needed) to restore the pre-change provider cooldown and gateway behavior.

## Validation

- [ ] `cargo test --workspace --locked`
- [ ] `cargo test --workspace --all-features --locked`

Commands and evidence:

```text
cargo fmt --all -- --check
cargo test -p loongclaw-app --lib
cargo test -p loongclaw

All three commands passed locally after the final fix set.

Before/after notes:
- Provider cooldowns now prefer provider hints even when shorter than the previous policy floor.
- Gateway compat accepted requests now route through shared runtime semantics instead of direct provider execution.
- Duplicate configured model ids are exposed as stable `profile_id:model` aliases.
- Streaming error responses now terminate with `[DONE]`.
- OpenAI-compatible requests that do not end in a `user` message are rejected instead of silently bypassing the shared runtime path.
```

- OpenAI-compatible clients can call `/v1/models` and `/v1/chat/completions` on the gateway and get shared-runtime behavior, usage propagation, stable duplicate-model aliases, and terminated SSE streams.
- Provider cooldown behavior now honors provider retry hints more precisely.

## Failure Recovery

  Revert the branch tip or disable use of the OpenAI-compatible gateway surface until the revert lands.
  Unexpected 400s for malformed/non-`user`-ending gateway chat payloads, incorrect model alias selection, missing usage in non-streaming compat responses, or channels no longer sending `"stream": true` in provider requests.

## Reviewer Focus

- Provider hint parsing and cooldown precedence in `crates/app/src/provider/rate_limit.rs` and `crates/app/src/provider/model_candidate_cooldown_runtime.rs`.
- Shared-runtime routing and usage propagation in `crates/daemon/src/gateway/openai_compat.rs`, `crates/app/src/agent_runtime.rs`, `crates/app/src/chat.rs`, and `crates/app/src/conversation/turn_coordinator.rs`.
- Channel streaming assertions in `crates/app/src/channel/feishu/webhook.rs`, `crates/app/src/channel/feishu/websocket.rs`, and `crates/app/src/channel/wecom.rs`.


gh-xj:
## LoongClaw QA Review — PR #1233

**Reviewed commit**: `77b73ddd`
**Risk**: high
**Agent**: `ai-scientist`

### Findings

- **high**: Streamed OpenAI-compatible turn execution can drop streamed tool calls because the streaming parser only reconstructs text deltas and finish markers, not streamed tool-call chunks. `crates/app/src/provider/contracts.rs:64`
- **medium**: The non-streaming OpenAI-compatible gateway flattens propagated upstream 4xx/429 failures into HTTP 500, which breaks client-visible retry/auth semantics. `crates/daemon/src/gateway/openai_compat.rs:229`

### Coverage

- Rust-specific review: applied (manual code-path review plus targeted Rust tests)
- Harness review: applied (repo-owned tests green; wrapper-path `loongclaw-dev` QA still lacks behavioral coverage)
- Adversarial challenge: applied (the `participant_id` QA build blocker did not reproduce under direct `cargo build`, so it remains a harness inconsistency rather than a confirmed branch defect)

### Open Questions

- GitHub currently reports this PR as `DIRTY`; rebase is required before closure.
- Wrapper-path E2E proof through the OpenAI-compatible gateway is still missing and should be rerun after rebase from a clean QA lane.

### Verdict

Blocking findings on the current head; rebase onto `dev`, fix the two runtime issues, and rerun boundary-aligned gateway QA before re-dispatch.

gh-xj (approved):

PR #1280, opened 2026-04-17
Fix feishu websocket fifo

### LINKED ISSUE #1273: [Bug]: Feishu websocket channel disconnects after a completed model turn and does not promptly reconnect
### Surface

Channels / integrations

### Severity

S1 - workflow blocked

### Regression?

Unknown

### Summary

When Feishu runs in websocket mode, the websocket connection is closed by the remote peer after a model turn completes. This is especially easy to hit on longer provider turns. After the close, the channel does not appear to reconnect promptly, so subsequent Feishu messages are not delivered until the process is restarted or a long delay passes.

This makes Feishu websocket mode impractical for normal interactive use.

### Expected behavior

A completed model turn should not cause the Feishu websocket session to be closed.

If the remote peer closes the websocket anyway, the Feishu channel runtime should reconnect automatically and promptly without requiring a manual restart.

Long-running turns should not block websocket heartbeats / reads in a way that causes the remote side to close the connection.

### Reproduction steps

```bash
1. Configure Feishu in websocket mode.
2. Start the Feishu channel runtime on the latest dev branch.
3. Send a Feishu message that causes a real model turn to run.
4. Wait for the model to finish and for Loong to send the reply.
5. Observe that the websocket session is closed by the remote peer after the turn.
6. Send another Feishu message.
7. Observe that the runtime does not promptly reconnect and the next message is not processed.
```

### Version / commit

dev@f9325b5e

### Operating system

Windows 11

### Runtime context

channel=feishu(websocket), provider=bailian_coding, model=glm-5

### Config / environment context

_No response_

### Logs / screenshots / evidence

```text

```

### Impact

_No response_

### Pre-flight checks

- [x] I reproduced this on the latest `dev` branch tip or the latest published release.
- [x] I redacted secrets, tokens, and private endpoints from everything I attached.
- [x] I searched existing issues and did not find the same report already tracked.

lihongtao1024:
This PR keeps the Feishu websocket session responsive to ping/pong traffic while preserving FIFO processing for data frames.

The previous background-processing approach fixed the ping/pong blocking issue, but it could reorder websocket replies because responses were sent in completion order. This change keeps the read loop non-blocking for control frames while ensuring websocket data-frame responses are still produced and sent in receive order.

## What Changed

- Replace concurrent `JoinSet`-style response completion ordering with a FIFO queue plus a single active task.
- Keep the websocket read loop free to continue handling:
  - control frames
  - ping interval ticks
  - newly received inbound frames
- Preserve ordered response delivery for websocket data frames.
- Keep deferred Feishu updates dispatched only after the corresponding websocket response is sent.
- Add regression coverage for ordered websocket reply handling.

## Why

Feishu websocket sessions need two properties at the same time:

1. Long-running provider turns must not block websocket ping/pong traffic.
2. Data-frame responses should remain FIFO to avoid reply reordering.

This change keeps both guarantees without reverting to a fully blocking per-frame processing model.

## Tests

Verified with:

- `cargo test -p loong-app feishu_websocket_session_preserves_reply_order_while_processing_in_background`
- `cargo test -p loong-app feishu_websocket_session_keeps_pinging_during_long_provider_turn`
- `cargo test -p loong-app feishu_websocket_session_reaches_provider_and_replies`

## Linked Issues

- Closes #1273

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [ ] Daemon / CLI / install
- [ ] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [x] Channels / integrations

## Risk Track

## Validation

- [x] Manually verified Feishu channel starts correctly via `cargo run --bin loong -- feishu-serve`.
- [x] Verified WebSocket connection survives after long turns.

Commands and evidence:

```bash
cargo run --bin loong -- feishu-serve
```

### DIFF
--- a/crates/app/src/channel/feishu/websocket.rs
+++ b/crates/app/src/channel/feishu/websocket.rs
@@ -1,5 +1,7 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, VecDeque};
+use std::future::{Future, pending};
 use std::path::Path;
+use std::pin::Pin;
 use std::sync::{Arc, OnceLock};
 use std::time::{Duration, Instant};
 
@@ -104,6 +106,13 @@ struct FeishuWsFragmentSet {
     parts: Vec<Option<Vec<u8>>>,
 }
 
+struct PendingWsDataFrame {
+    payload_bytes: Vec<u8>,
+    frame: FeishuWsFrame,
+}
+
+type ActiveWsTask = Pin<Box<dyn Future<Output = Result<FeishuWsCompletedTask, String>> + Send>>;
+
 #[derive(Serialize)]
 struct FeishuWsResponseEnvelope {
     code: u16,
@@ -203,14 +212,14 @@ pub(super) async fn run_feishu_websocket_channel(
 ) -> CliResult<()> {
     let mut adapter = FeishuAdapter::new(resolved)?;
     adapter.refresh_tenant_token().await?;
-    let state = FeishuWebhookState::new_with_resolved_path(
+    let state = Arc::new(FeishuWebhookState::new_with_resolved_path(
         config.clone(),
         resolved_path.to_path_buf(),
         resolved,
         adapter,
         kernel_ctx,
         runtime,
-    );
+    ));
     let client = FeishuClient::from_configs(resolved, &config.feishu_integration)?;
 
     #[allow(clippy::print_stdout)]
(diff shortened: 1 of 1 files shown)


chumyin (approved):
hey @lihongtao1024, thanks for pushing this through — i rebased the websocket fix onto current dev, kept the session loop responsive for ping/pong traffic, preserved fifo reply ordering, and added regression coverage for both the long-turn heartbeat path and the ordered reply path. i also dropped the stale daemon runtime change because current dev already has the safer runtime stack handling in place. nice catch, and congrats on landing a solid fix.

PR #1296, opened 2026-04-18
Refactor session-core durability away from memory-augmentation surfaces

chumyin:
## Summary

This change finishes the current convergence lane for session durability and memory augmentation without changing the current crate graph.

The main outcome is that session transcript/store semantics now have one explicit home under `session::store`, while true memory-augmentation paths stay on the memory side.

## What changed

- added `crates/app/src/session/store.rs` as the session-core front door for:
  - transcript append
  - transcript window reads
  - transcript paging
  - session-store config construction

- migrated session-core callers to that surface across:
  - chat / control-plane session views
  - session tools
  - session search
  - messaging
  - work repository
  - conversation runtime / turn loop / turn engine / turn coordinator
  - operator session and delegate helpers

- updated large conversation test fixtures to use session-store helpers instead of direct `memory::*` transcript calls

- added memory-side helper cleanup for augmentation-facing paths:
  - `MemoryRuntimeConfig::for_sqlite_path(...)`
  - canonical-memory sqlite-path helpers
  - stage-envelope request / hydration helpers for memory config

- documented the intended boundary in `docs/design-docs/single-entry-runtime-convergence.md`

## Why

Before this refactor, session durability and transcript operations were semantically mixed into `memory::*`, which made session-core code look like an optional memory feature.

After this change:

- session transcript/store code reads as session-core
- memory augmentation code reads as augmentation
- remaining `memory::*` references are intentional:
  - the thin `session::store` adapter
  - augmentation paths such as staged envelope / hydration
  - memory-facing tools and their tests

## Validation

## Risk

Moderate. The diff is wide, but behavior is preserved and the work is mostly ownership/entry-surface refactoring plus test fixture alignment.

## Follow-up

A separate follow-up can further refine memory-augmentation architecture, but this PR completes the session-core convergence lane.

### DIFF
--- a/crates/app/src/acp/store.rs
+++ b/crates/app/src/acp/store.rs
@@ -122,10 +122,11 @@ impl AcpSqliteSessionStore {
     }
 
     fn resolved_path(&self) -> CliResult<PathBuf> {
-        let runtime_config = crate::memory::runtime_config::MemoryRuntimeConfig {
-            sqlite_path: self.path.clone(),
-            ..crate::memory::runtime_config::MemoryRuntimeConfig::default()
-        };
+        let runtime_config = self
+            .path
+            .clone()
+            .map(crate::memory::runtime_config::MemoryRuntimeConfig::for_sqlite_path)
+            .unwrap_or_default();
         crate::memory::ensure_memory_db_ready(self.path.clone(), &runtime_config)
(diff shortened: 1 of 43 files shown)


gh-xj:
Review started. I'll post findings here when ready.

PR #1306, opened 2026-04-19
Refactor session-core durability away from memory augmentation surfaces

### LINKED ISSUE #1303: Clarify session-core ownership away from memory augmentation naming
## Problem

Session transcript and durability code still reads as if it belongs to the optional memory subsystem, even though it is product-core behavior.

That makes ownership harder to reason about across chat, conversation runtime, tools, and session repository surfaces.

## Desired outcome

- give session transcript and durability paths one explicit front door
- keep true memory augmentation on the memory side
- preserve the current 7-crate DAG and runtime behavior

## Scope

This issue covers:

- introducing and adopting a `session::store` front door
- moving session-facing call sites to that surface
- cleaning up remaining session-facing naming leaks where `SessionStoreConfig` still reads as `MemoryRuntimeConfig`

This issue does not cover:

- splitting new crates
- redesigning memory augmentation itself
- changing user-facing runtime behavior

chumyin:
## Summary

This PR gives session transcript and durability paths one explicit home under `session::store` while keeping true memory augmentation logic on the memory side.

## What changed

- added `crates/app/src/session/store.rs` as the session-core front door for transcript and durability primitives
- migrated session-facing callers across chat, conversation runtime, session tools, repositories, and operator paths to that surface
- clarified the intended boundary in `docs/design-docs/single-entry-runtime-convergence.md`
- removed the last session-facing naming leaks where `SessionStoreConfig` still read as `MemoryRuntimeConfig`

## Why

Before this change, session durability and transcript operations were semantically mixed into `memory::*`, which made product-core session behavior look like an optional memory feature.

After this change:

- session transcript/store code reads as session-core
- true augmentation code stays on the memory side
- the current crate contract stays intact

## Validation

Closes #1303

### DIFF
--- a/crates/app/src/acp/store.rs
+++ b/crates/app/src/acp/store.rs
@@ -122,10 +122,11 @@ impl AcpSqliteSessionStore {
     }
 
     fn resolved_path(&self) -> CliResult<PathBuf> {
-        let runtime_config = crate::memory::runtime_config::MemoryRuntimeConfig {
-            sqlite_path: self.path.clone(),
-            ..crate::memory::runtime_config::MemoryRuntimeConfig::default()
-        };
+        let runtime_config = self
+            .path
+            .clone()
+            .map(crate::memory::runtime_config::MemoryRuntimeConfig::for_sqlite_path)
+            .unwrap_or_default();
         crate::memory::ensure_memory_db_ready(self.path.clone(), &runtime_config)
             .map_err(|error| format!("resolve ACP sqlite store path failed: {error}"))
     }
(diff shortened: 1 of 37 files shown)


chumyin:
closing this because origin/dev already contains the newer session-store convergence in fdc0b4c44 and ef3f17e9f. this branch now conflicts with current dev and would only re-introduce an older variant of the same work, so keeping the already-landed path is the cleaner option.

chumyin:
closing this as stale. the session::store convergence already landed on dev in fdc0b4c44 and ef3f17e9f, so this branch no longer carries a distinct reviewable delta. the remaining sqlite-backed test cleanup continues in #1307.

PR #1339, opened 2026-04-20
Skip installer legacy cleanup when loongclaw is not a file

chumyin:
## Summary

- Problem: the stale `loongclaw` cleanup added in #1332 removes any same-name path in the install prefix, so the bash installer now fails if `loongclaw` is a directory instead of an old executable.
- Why it matters: the cleanup is meant to narrow the shipped command surface, not mutate arbitrary same-name paths or turn an upgrade into a partial install failure.
- What changed: both installers now only remove stale legacy executable artifacts, and the shell installer regression suite now covers the directory-collision case.
- What did not change (scope boundary): the single-binary `loong` delivery contract, the stale executable cleanup itself, and the rest of the install prefix remain unchanged.

## Linked Issues

- Closes #
- Related #1332

## Change Type

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [x] Daemon / CLI / install
- [ ] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [ ] Channels / integrations

## Risk Track

If Track B, fill these in:

- Risk notes:
- Rollout / guardrails:
- Rollback path:

## Validation

- [x] Relevant architecture / dep-graph / docs checks for touched areas

Commands and evidence:

```text
bash -n scripts/install.sh
bash -n scripts/test_install_sh.sh
bash scripts/test_install_sh.sh
bash scripts/test_release_workflow_hardening.sh
LOONG_ARCH_STRICT=true ./scripts/check_architecture_boundaries.sh
./scripts/cargo-local-toolchain.sh fmt --all -- --check
./scripts/cargo-local-toolchain.sh clippy --workspace --all-targets --all-features -- -D warnings
./scripts/cargo-local-toolchain.sh test --workspace --locked
./scripts/cargo-local-toolchain.sh test --workspace --all-features --locked
```

Additional behavior check:
- manually reproduced the pre-fix failure by creating `<prefix>/loongclaw/` as a directory before install and verified the fixed installer now leaves that directory alone while still installing `loong` successfully.
- local PowerShell execution was not available, so `scripts/install.ps1` was validated by code review rather than native runtime execution.

- Re-running the installers still removes stale `loongclaw` executable artifacts.
- Installer runs no longer fail when the install prefix already contains a directory named `loongclaw` / `loongclaw.exe`.

## Failure Recovery

- Fast rollback or disable path: revert the file-type guard in `scripts/install.sh` and the container guard in `scripts/install.ps1`, then remove the new directory-collision regression test.
- Observable failure symptoms reviewers should watch for: installer cleanup starts deleting non-file paths again, or stale executable aliases stop being removed.

## Reviewer Focus

- `scripts/install.sh` for the file-vs-directory guard around legacy cleanup.
- `scripts/install.ps1` for the equivalent literal-path and container check.
- `scripts/test_install_sh.sh` for the new directory-collision regression coverage.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -105,11 +105,15 @@ function Install-Binary([string]$SourceBinary) {
 function Remove-LegacyBinaryIfPresent {
     $legacyBinaryName = "loongclaw.exe"
     $legacyBinary = Join-Path $Prefix $legacyBinaryName
-    if (-not (Test-Path $legacyBinary)) {
+    $legacyBinaryItem = Get-Item -LiteralPath $legacyBinary -Force -ErrorAction SilentlyContinue
+    if ($null -eq $legacyBinaryItem) {
+        return
+    }
+    if ($legacyBinaryItem.PSIsContainer) {
         return
     }
 
-    Remove-Item -Force $legacyBinary
+    Remove-Item -LiteralPath $legacyBinary -Force
     Write-Host "==> Removed legacy loongclaw compatibility command from $legacyBinary"
 }
 
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -363,15 +363,14 @@ install_binary() {
 remove_legacy_binary_if_present() {
   local legacy_output_name="${1:?legacy_output_name is required}"
   local legacy_output_path="${prefix}/${legacy_output_name}"
-  local legacy_output_exists=0
 
-  if [[ -e "${legacy_output_path}" ]]; then
-    legacy_output_exists=1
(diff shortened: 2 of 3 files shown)


chumyin:
quick follow-up for #1332: this narrows the legacy cleanup to files and symlinks only, keeps same-name directories untouched, and adds regression coverage for the directory-collision path. local install, governance, architecture, and cargo checks all passed.

PR #1343, opened 2026-04-20
(fix)feishu: surface retry progress and inline provider failures

### LINKED ISSUE #1281: [Bug]: Feishu websocket redelivers timed-out streaming turns indefinitely and logs streaming retry metadata as 1/1
### Surface

Channels / integrations

### Severity

S2 - degraded behavior

### Regression?

Unknown

### Summary

When Loong runs a streamed provider turn from the Feishu websocket channel and the provider request times out during SSE body reading, the failure is not surfaced clearly at the channel level and the same Feishu event appears to be processed repeatedly.

There are two related defects:
1. A provider timeout in the streaming path causes Feishu inbound handling to fail in a way that allows the same event to be redelivered and processed again, which looks like an infinite retry loop from the operator side. This redelivery behavior is also not bounded by the provider retry_max_attempts setting.
2. The streaming timeout path rewrites retry metadata in logs so the emitted provider failure snapshot always shows attempt=1 max_attempts=1, even when the configured provider retry budget is higher.

In practice this makes a simple timeout look like an endless loop with misleading retry metadata.

### Expected behavior

* A streamed provider timeout should be surfaced explicitly as a streaming timeout or stream body read timeout, not only as the generic error decoding response body.
* Feishu inbound processing should not repeatedly re-run the same timed-out event in a way that looks unbounded from the operator side.
* If the repeated executions are caused by channel redelivery rather than provider retry, logs should make that distinction obvious.
* Streaming timeout errors should preserve the real provider retry context, including the actual attempt, max_attempts, and model information.
* The observed retry metadata in logs should match the configured provider retry policy instead of being hard-coded to 1/1.

### Reproduction steps

```bash
1. Configure a Feishu websocket channel.
2. Configure a streaming provider such as `bailian_coding`.
3. Set a small provider timeout, for example:
   [provider]
   request_timeout_ms = 30000
   retry_max_attempts = 3
4. Send a prompt that causes the model to respond slowly enough that the streaming body read times out.
5. Observe:
   - the provider request returns HTTP 200 with `content-type: text/event-stream`
   - the stream later fails with `operation timed out`
   - Feishu processing re-enters `request_turn_streaming` for the same inbound flow
   - provider failure logs report `attempt=1 max_attempts=1` instead of the configured retry budget
```

### Version / commit

dev@latest

### Operating system

Windows 11

### Runtime context

channel=feishu websocket, provider=bailian, model=glm-5

### Config / environment context

[provider]
kind = "bailian"
request_timeout_ms = 30000
retry_max_attempts = 3

[channels.feishu]
mode = "websocket"

### Logs / screenshots / evidence

```text

```

### Impact

_No response_

### Pre-flight checks

- [x] I reproduced this on the latest `dev` branch tip or the latest published release.
- [x] I redacted secrets, tokens, and private endpoints from everything I attached.
- [x] I searched existing issues and did not find the same report already tracked.

lihongtao1024:
## Summary

- Problem:
  Feishu inbound handling had two linked gaps around provider retry behavior. First, streaming provider requests could fail before producing any usable stream output without consuming the configured retry budget. Second, when provider execution ultimately failed, Feishu users could end up with no visible progress or error feedback.
- Why it matters:
  From the user side this looked like a silent bot stall. From the runtime side it made streaming behavior less robust than non-streaming behavior and hid retry progress that would otherwise explain the delay.

- What changed:
  Added provider retry progress propagation for turn requests, taught streaming turn execution to safely retry transport failures before the first delivered stream event, and added a Feishu text status-message MVP that creates one status reply on first retry, updates it across retries, and updates it again with a user-facing error on terminal failure. Feishu inbound final provider errors now use inline message mode so failures are surfaced back into the conversation.
  This does not add card-based retry status UI, does not change global provider retry policy semantics, does not disable streaming for channel flows, and does not alter non-Feishu channel UX.

## Linked Issues

- Closes #1281

## Change Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [ ] Daemon / CLI / install
- [x] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [x] Channels / integrations

## Risk Track

## Validation

Commands and evidence:

```text
cargo check -p loong-app
Result: passed

cargo check -p loong-app --tests
Result: passed

cargo fmt --all --check
Result: passed

cargo test -p loong-app --lib channel::feishu::webhook::tests::feishu_webhook_provider_timeout_acknowledges_after_retry_budget_exhaustion -- --exact --nocapture
Result: passed

cargo test -p loong-app --lib channel::feishu::webhook::tests::feishu_retry_status_handle_creates_and_updates_single_status_message -- --exact --nocapture
Result: passed

cargo test -p loong-app --lib provider::request_executor::tests::execute_streaming_turn_request_retries_timeout_before_first_event -- --exact --nocapture
Result: passed

cargo test -p loong-app --lib provider::request_dispatch_runtime::tests::request_turn_streaming_with_transport_preserves_timeout_context_for_stream_errors -- --exact --nocapture
Result: passed
```

- Feishu users now see a single retry status message when provider retries begin instead of waiting through a silent stall.
- If retries recover, the status message updates to indicate recovery before the final answer is sent.
- If retries are exhausted, Feishu users receive a user-facing failure update instead of only an internal webhook ack path.
- Streaming requests now retry transport failures before the first delivered stream event, improving robustness without duplicating already-emitted output.

## Failure Recovery

  Revert this PR, or locally remove Feishu inline provider error + retry progress wiring by reverting the `process_inbound_with_provider_and_error_mode_and_retry_progress(...)` path in Feishu webhook handling.
  Duplicate Feishu status messages instead of a single updated one, missing final answer after a recovered retry, or missing final error update when retries are exhausted.

## Reviewer Focus

- Review [request_executor.rs](/h:/gitee/loong/crates/app/src/provider/request_executor.rs) around the streaming retry boundary:
  only transport failures before the first delivered stream event should retry.
- Review [webhook.rs](/h:/gitee/loong/crates/app/src/channel/feishu/webhook.rs) around the Feishu status message lifecycle:
  first retry creates one status reply, later retries update it, final success/failure closes it out.
- Review [dispatch.rs](/h:/gitee/loong/crates/app/src/channel/dispatch.rs), [mod.rs](/h:/gitee/loong/crates/app/src/provider/mod.rs), and [runtime.rs](/h:/gitee/loong/crates/app/src/conversation/runtime.rs) for the retry-progress callback threading:
  the callback should stay optional and should not change unrelated channel or conversation flows.
- Pay special attention to the scope boundary between provider retry internals and Feishu UX:
  provider retry signaling should remain transport-aware, while Feishu rendering should stay user-facing and avoid leaking raw `[provider_error]` envelopes.


chumyin (approved):
hi @lihongtao1024, thanks for the solid fix here. i added a small generic follow-up on top to keep streaming timeout diagnostics accurate after response headers arrive, and the updated branch is green now. approving and landing.

PR #1401, opened 2026-04-23
docs(README): call out CN Feishu vs international Lark domain in onboard

Binlogo:
## Summary

- Document the `--domain feishu` vs `--domain lark` split in the Feishu onboarding section of both `README.md` and `README.zh-CN.md` so CN users do not silently land on `open.larksuite.com`.
- Note that omitting `--domain` falls back to the CLI default (`feishu`), matching `FeishuOnboardDomainArg::Feishu` in `crates/daemon/src/feishu_cli.rs`.

## Why

The prior README only showed `loong feishu onboard --domain lark` with a single toml comment hinting at the CN lane. Tracking eastreams/loong#1348 — CN users reported running the example verbatim and getting routed to the international registration endpoint.

## Test plan

- [x] README renders both command variants
- [x] Diff limited to README.md and README.zh-CN.md — no code paths touched

Refs eastreams/loong#1348

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -178,8 +178,14 @@ model = "auto"
 推荐先走命令行二维码接入：
 
 ```bash
-loong feishu onboard
+# 国内飞书租户（open.feishu.cn）—— CLI 默认值
+loong feishu onboard --domain feishu
+
+# 国际版 Lark 租户（open.larksuite.com）
+loong feishu onboard --domain lark
 ```
+
+`--domain` 一定要和你登录的租户类型对上：`feishu` 走国内 `open.feishu.cn`，`lark` 走国际 `open.larksuite.com`。选错端点时二维码扫完也绑不上对应 app。省略 `--domain` 时默认走 `feishu`。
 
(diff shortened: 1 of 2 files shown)


README.zh-CN.md
```diff
@@ -178,9 +178,15 @@ model = "auto"
 推荐先走命令行二维码接入：
 
 ```bash
-loong feishu onboard
+# 国内飞书租户（open.feishu.cn）—— CLI 默认值
```
chumyin, line 181: `租户` might be a bit too 机翻😂, recommend change it to `国内飞书用户` or just simply remove `租户`

chumyin (changes requested):

README.zh-CN.md
Binlogo: "租户" is actually a more precise term in a strict sense. This is also how it is expressed on Lark's official open platform.

chumyin:
discussed with @Binlogo on wechat, made common sense and agreed that `租户` is actually more precise, so this PR is good to go. thanks so much for the contribution here!

README.zh-CN.md
chumyin: agreed, let's keep this!

chumyin (approved):

PR #1405, opened 2026-04-25
fix(memory): prevent session-tree rewrite from orphaning non-active heads and artifacts

### LINKED ISSUE #1406: Bug: context-compaction can silently orphan named session-tree heads / artifacts
## Affected

`dev` branch, observed on commit `a0bbf0e` (PR #1383 landed 2026-04-24).

## Production trigger

`crates/app/src/conversation/compaction.rs:32-54` -- `compact_window` is reachable from `context_engine.rs:405` (`maybe_persist_compacted_memory_window` -> `persist_memory_window` -> `build_replace_turns_request_with_expectation` -> `replace_turns_internal`). Its output length is strictly `preserve + 1` (one summary turn plus the preserved tail), which is **strictly less than `turns.len()` whenever `older.len() > 1`**. Concrete example: 10 turns with `preserve_recent_turns = 5` produces 6 turns -- the four oldest turns are replaced with a single summary.

```rust
pub fn compact_window(turns: &[WindowTurn], policy: CompactPolicy) -> Option<Vec<WindowTurn>> {
    let preserve = policy.preserve_recent_turns.min(turns.len());
    if turns.len() <= preserve { return None; }
    let split_at = turns.len() - preserve;
    let (older, recent) = turns.split_at(split_at);
    ...
    let mut compacted = Vec::with_capacity(recent.len() + 1);
    compacted.push(summary);
    compacted.extend_from_slice(recent);
    Some(compacted)
}
```

Any long-running session that hits the context-window policy exercises this path -- routine, not pathological.

## Symptom

`rebuild_linear_session_tree_for_turns` (`crates/app/src/memory/sqlite.rs:1649-1769`) deletes every `session_nodes` row for the session, re-inserts only the new turn nodes, and upserts only the `active` head. Non-`active` heads (e.g. `checkpoint/foo`, `thread/alpha`) and `session_artifacts` rows keep their `node_id` / `anchor_node_id` / `source_start_node_id` / `source_end_node_id` pointing at deleted node IDs. There is no FK cascade (the three session-tree tables in `ensure_session_tree_storage` at `crates/app/src/memory/sqlite.rs:2557-2598` carry no `FOREIGN KEY` clauses despite `PRAGMA foreign_keys = ON` being set per-connection at `:2364`), no trigger, and no post-rebuild sweep.

The reader silences the failure: `load_session_path_for_head_with_conn` (`crates/app/src/session/repository.rs:936-938`) `break`s on a missing node and returns `Ok(Vec::new())`, so a dangling checkpoint is indistinguishable from an empty one at the operator surface. (The adjacent `head-missing` early return at `:929-931` is defensible -- a session that never forked a head should indeed produce an empty path. It is the **mid-walk** silent break that masks corruption.)

The three tool-level callers of `load_session_path_for_head` handle the empty path inconsistently, which makes the silent mode harder to spot in practice:

| Caller | Behaviour on empty path |
|---|---|
| `execute_session_path` (`crates/app/src/tools/session.rs:1928-1940`) | success payload, `node_count: 0` |
| `execute_session_create_branch_summary` (`crates/app/src/tools/session.rs:2466-2469`) | hard error `"session head \`...\` not found"` -- accidentally surfaces the bug, with a misleading message |
| `load_session_tree_snapshot_record_with_conn` (`crates/app/src/tools/session.rs:3279`) | only reads `active`, which `rebuild` always repairs -- not affected |

PR #1383's body explicitly listed "metadata loss during transcript rewrites" as a risk the durable tree layer was meant to address; silent orphaning of named heads during rewrite is the literal failure mode the durable tree was supposed to prevent.

## Why the existing regression test misses it

`replace_turns_preserves_named_heads_and_artifacts_across_rewrites` (`crates/app/src/session/repository.rs:4998-5070`) rewrites 2 turns -> 2 turns. The fork target is `session-turn:{sid}:1`; because turn-node IDs are deterministic by index, the rebuild re-creates a node with the same ID and the head's `node_id` pointer happens to still resolve. The test never exercises a length change.

## Mutation-side fix

PR #1405 (link added once filed) implements the mutation-side fix, mirroring jujutsu's `MutableRepo::update_rewritten_references` (`jj-vcs/jj`, `lib/src/repo.rs:1142`). Inside the same transaction that rebuilds the tree, non-`active` heads with out-of-range targets are dropped (with a `session_tree_rewrite_dropped_head` event capturing the pre-rewrite content); artifacts have their out-of-range node-id columns nulled (their `payload_json` / `summary_text` survive, so the artifact's own content is preserved); and `artifact.head_name` for dropped heads is also nulled to keep the two tables referentially consistent.

The PR carries a regression test that fails on current `dev` and passes after the fix, providing concrete RED -> GREEN evidence.

## Adjacent follow-ups (defence-in-depth, separate from PR #1405)

These are not blocking and not addressed in PR #1405 -- after the mutation-side fix lands, the data can no longer reach a dangling state via the rewrite path. They are still worth doing as defence-in-depth.

1. **Declare FK constraints.** `session_heads.node_id` and `session_artifacts.{anchor,source_start,source_end}_node_id` should carry real `FOREIGN KEY` clauses against `session_nodes(node_id)`; the four tables should also FK onto `sessions(session_id)`. The pragma is already set per-connection -- only the schema declarations are missing. Reference shape: Zed (`zed-industries/zed`, `crates/db/src/db.rs:125-127`).
2. **Distinguish "absent" from "dangling" in the reader.** Change `load_session_path_for_head_with_conn` to return a typed `Err(LoadSessionPathError::DanglingHead { head_name, missing_node_id })` on mid-walk node-miss; keep `Ok(Vec::new())` only for the legitimate "no such head" early return at `:929-931`. Mirrors `gix_ref::peel::to_id::Error::NotFound { oid, name }` (`GitoxideLabs/gitoxide`, `gix-ref/src/peel.rs:14`) and sqlx's `MigrateError::VersionMissing` vs `VersionNotPresent` split (`launchbadge/sqlx`, `sqlx-core/src/migrate/error.rs:16,22`). The contract change is real but limited to three operator-tool call sites in `crates/app/src/tools/session.rs`.

## Minor observations

Logged for record, low priority compared to the above:

- **`set_session_head` silently preserves `Pinned` mode** when relocating a non-active head (`crates/app/src/session/repository.rs:1644-1678`). Defensible either way, but worth documenting in the `session_pin_head` tool description so the chosen semantics is discoverable.
- **`append_turn_session_tree_node` uses `INSERT OR IGNORE`** (`crates/app/src/memory/sqlite.rs:1564`, function at `:1541`). If the deterministic turn-node ID ever collides (upstream turn-index non-monotonicity), the insert silently no-ops while the `active` head is still advanced. Plain `INSERT` would fail loudly when that invariant breaks.
- **`Children` visibility + `search_scope = lineage_family`** broadens to sibling / cousin sessions by design (the opt-in is the explicit scope field). `SelfOnly` visibility remains unconditionally clamped to `[current_session_id]` at `crates/app/src/tools/session_search.rs:109-111`, so the SelfOnly combo is safe. Flagging only so operators who use `Children` as a sandbox (not just a usability default) know it is escapable through the scope field.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Mike-7777777:
## Summary

`rebuild_linear_session_tree_for_turns` (`crates/app/src/memory/sqlite.rs:1649-1769`) deletes every `session_nodes` row for the session and re-inserts only the new turn nodes, then upserts only the `active` head. When the new transcript is **shorter** than the original — e.g. context-window compaction (`crates/app/src/conversation/compaction.rs:32-54` is provably output-shorter when `older.len() > 1`) — non-`active` heads and artifacts that referenced turn nodes past the new tail are left pointing at deleted node IDs. The pointers then resolve to `Ok(Vec::new())` via `load_session_path_for_head_with_conn` (`crates/app/src/session/repository.rs:936-938`), so a dangling checkpoint is indistinguishable from an empty one at the operator surface.

PR #1383 explicitly listed "metadata loss during transcript rewrites" as a risk the durable tree layer was meant to prevent — this was the literal failure mode.

## The fix (mutation-side, jj pattern)

Mirror jujutsu's `MutableRepo::update_rewritten_references` (`jj-vcs/jj`, `lib/src/repo.rs:1142`): inside the same transaction that rebuilds the tree, walk non-`active` heads and artifacts whose target turn-index falls past the new tail (or whose target is already missing — pre-existing legacy corruption), and:

- **Drop stale heads.** Emit a `session_tree_rewrite_dropped_head` `session_events` row capturing the original `head_name`, `stale_node_id`, and a snapshot of the original node's `content`, so operators can audit what was lost.
- **Null out-of-range artifact node refs.** `anchor_node_id` / `source_start_node_id` / `source_end_node_id` are set to NULL on a per-column basis (only the columns that actually dangle). The artifact's own `payload_json` and `summary_text` are not node-referential and survive intact -- the artifact's content is preserved. Emit a `session_tree_rewrite_nulled_artifact_refs` event recording the original IDs.
- **Null `artifact.head_name`** when the referenced head was just dropped, keeping referential consistency between the two tables.

Implementation lives in a new `preserve_session_tree_before_rebuild` helper in `crates/app/src/memory/sqlite.rs`, called from `rebuild_linear_session_tree_for_turns` before the existing `DELETE FROM session_nodes`. Net diff: +207 lines in `sqlite.rs`, +130 lines in the new test in `repository.rs`.

## Test evidence (RED -> GREEN)

Six new regression tests in the `tests` mod of `crates/app/src/session/repository.rs`, covering the primary scenario plus five edge cases that an independent review surfaced. The pre-existing `replace_turns_preserves_named_heads_and_artifacts_across_rewrites` (same-length rewrite) still passes unchanged.

| Test | Scenario | Why it matters |
|------|----------|----------------|
| `replace_turns_shorter_drops_stale_head_and_nulls_artifact_refs` | 5 turns -> 2 turns; head + artifact anchored at turn 5 | Primary bug repro: head dropped, artifact refs nulled, audit events recorded with content snapshot |
| `replace_turns_shorter_only_nulls_dangling_artifact_columns` | Artifact spans turns 1-5; rewrite to 2 turns | Per-column null logic: anchor + source_start (turn 1) preserved, source_end (turn 5) nulled |
| `replace_turns_shorter_drops_multiple_stale_heads_with_separate_events` | Three heads at turns 5 / 4 / 1; rewrite to 2 turns | Two stale heads dropped with separate audit events; in-range head survives |
| `replace_turns_empty_drops_non_active_heads_and_keeps_root` | `replace_turns(&[])` (zero turns) | Active head repoints at root; non-active heads dropped; root node preserved via deterministic id |
| `replace_turns_cleans_up_legacy_dangling_head` | Pre-existing head whose target node was already missing | Preservation phase also cleans up legacy corruption that predates this PR (`n.node_id IS NULL` branch) |
| `replace_turns_same_length_emits_no_preservation_events` | 2 turns -> 2 turns with head + artifact at turn 1 | No-op assertion: zero `session_tree_rewrite_*` events when nothing dangles |

Local verification on `dev @ a0bbf0e` + this branch:

```
Stage 0a (fmt --check):           exit=0   PASSED
Stage 0b (clippy -D warnings):    exit=0   PASSED
Stage 1  (RED, test without fix): exit=101 FAILED as expected, bug reproduces
Stage 2  (GREEN, test with fix):  exit=0   PASSED, fix works
Stage 3  (SANITY, existing test): exit=0   PASSED, prior contract preserved
```

Run reproducible via `cargo test -p loong-app replace_turns_ --locked` (matches all six new tests + the pre-existing same-length test).

Update on 2026-04-30: tightened `replace_turns_shorter_drops_stale_head_and_nulls_artifact_refs` to also assert `artifact.head_name` is nulled in lockstep with the dropped head — closes a coverage gap where the `head_name`-nulling UPDATE could regress silently.

## Scope and intentional non-coverage

Two adjacent fixes are intentionally out of this PR -- they belong to a separate Issue (will link in a follow-up comment):

- **Schema-layer FK declarations.** `PRAGMA foreign_keys = ON` is set per-connection in `configure_sqlite_connection` (`crates/app/src/memory/sqlite.rs:2360-2370`) but `session_nodes` / `session_heads` / `session_artifacts` carry no `FOREIGN KEY` clauses, so the pragma has nothing to enforce. Adding them would be defence-in-depth on top of this PR's mutation-side fix and matches the per-connection pragma + declared FKs shape used elsewhere in the rusqlite ecosystem (e.g. `zed-industries/zed`, `crates/db/src/db.rs:125-127`). Schema migration is non-trivial because SQLite cannot `ALTER TABLE ADD FOREIGN KEY`, so it deserves its own commit.
- **Reader-side error variant.** `load_session_path_for_head_with_conn` collapses "head absent" and "head resolves but target missing" into the same `Ok(Vec::new())` return; the latter case is corruption being masked. Splitting it into a typed error (`DanglingHead { head_name, missing_node_id }`) would mirror `gix_ref::peel::to_id::Error::NotFound { oid, name }` (`GitoxideLabs/gitoxide`, `gix-ref/src/peel.rs:14`) and sqlx's `MigrateError::VersionMissing` vs `VersionNotPresent` split (`launchbadge/sqlx`, `sqlx-core/src/migrate/error.rs:16,22`). Touches the `Result` type of three call sites in `crates/app/src/tools/session.rs` and is a separate API change.

After this PR's mutation fix lands, both are pure defence-in-depth -- the data can no longer reach a dangling state via the rewrite path.

## Verification commands

```
cargo fmt --all -- --check
cargo clippy -p loong-app --all-targets --all-features --locked -- -D warnings
cargo test -p loong-app replace_turns_ --locked     # 6 new + 1 sanity, all pass
```

All four pass on `stable-x86_64-pc-windows-gnu` + a separately installed full MinGW-w64. (Windows-GNU also needs `-A dead_code -A unused_imports -A unused_variables` on clippy to suppress `#[cfg(unix)]`-gated false positives that the upstream Linux CI does not see; not relevant under the project's actual CI.)

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Fixes #1406

### DIFF
--- a/crates/app/src/memory/sqlite.rs
+++ b/crates/app/src/memory/sqlite.rs
@@ -1646,6 +1646,197 @@ fn append_turn_session_tree_node(
     Ok(())
 }
 
+fn preserve_session_tree_before_rebuild(
+    conn: &Connection,
+    session_id: &str,
+    turns: &[WindowTurn],
+    fallback_ts: i64,
+) -> Result<(), String> {
+    let new_tail_turn_index: i64 = turns.len() as i64;
+    let preservation_ts: i64 = turns.last().and_then(|turn| turn.ts).unwrap_or(fallback_ts);
+
+    // --- Drop non-active heads whose target would dangle.
+    //
+    // Includes: (a) heads pointing at a node past the new tail, (b) heads
+    // pointing at a node that's already missing (legacy corruption).
+    // The session root node (session_turn_index IS NULL) is always safe —
+    // the rebuild below re-creates it.
+    let stale_heads: Vec<(String, String, Option<String>)> = {
+        let mut stmt = conn
+            .prepare(
+                "SELECT h.head_name, h.node_id, n.content
+                 FROM session_heads h
+                 LEFT JOIN session_nodes n ON n.node_id = h.node_id
+                 WHERE h.session_id = ?1
+                   AND h.head_name != ?2
+                   AND (
+                        n.node_id IS NULL
+                        OR (n.session_turn_index IS NOT NULL
+                            AND n.session_turn_index > ?3)
+                   )",
+            )
+            .map_err(|error| format!("prepare stale head query failed: {error}"))?;
+        let rows = stmt
+            .query_map(
+                rusqlite::params![session_id, ACTIVE_SESSION_HEAD_NAME, new_tail_turn_index],
+                |row| {
+                    Ok((
+                        row.get::<_, String>(0)?,
+                        row.get::<_, String>(1)?,
+                        row.get::<_, Option<String>>(2)?,
+                    ))
+                },
+            )
+            .map_err(|error| format!("scan stale heads failed: {error}"))?;
+        let mut out = Vec::new();
+        for row in rows {
+            out.push(row.map_err(|error| format!("read stale head row failed: {error}"))?);
+        }
+        out
+    };
+
+    for (head_name, stale_node_id, content_snapshot) in &stale_heads {
+        let payload = json!({
+            "head_name": head_name,
+            "stale_node_id": stale_node_id,
+            "new_tail_turn_index": new_tail_turn_index,
+            "content_snapshot": content_snapshot,
+        });
+        conn.execute(
+            "INSERT INTO session_events(
+                session_id, event_kind, actor_session_id, payload_json, search_text, ts
+             ) VALUES (?1, 'session_tree_rewrite_dropped_head', NULL, ?2, '', ?3)",
+            rusqlite::params![session_id, payload.to_string(), preservation_ts],
+        )
+        .map_err(|error| format!("record dropped head event failed: {error}"))?;
+        conn.execute(
+            "DELETE FROM session_heads
+             WHERE session_id = ?1 AND head_name = ?2",
+            rusqlite::params![session_id, head_name],
+        )
+        .map_err(|error| format!("drop stale session head failed: {error}"))?;
+    }
+
+    // --- Null dangling *_node_id columns on artifacts.
+    //
+    // The artifact's own payload_json / summary_text is not node-referential
+    // and is left intact — only the back-references are cleared.
+    let stale_artifact_refs: Vec<(
+        String,
+        Option<String>, // current anchor_node_id
+        Option<String>, // current source_start_node_id
+        Option<String>, // current source_end_node_id
+        Option<i64>,    // anchor target's session_turn_index (NULL when absent)
+        Option<i64>,    // source_start target's session_turn_index
+        Option<i64>,    // source_end target's session_turn_index
+        bool,           // anchor exists in session_nodes
+        bool,           // source_start exists in session_nodes
+        bool,           // source_end exists in session_nodes
+    )> = {
+        let mut stmt = conn
+            .prepare(
+                "SELECT a.artifact_id,
+                        a.anchor_node_id,       a.source_start_node_id,    a.source_end_node_id,
+                        na.session_turn_index,  nss.session_turn_index,    nse.session_turn_index,
+                        na.node_id IS NOT NULL, nss.node_id IS NOT NULL,   nse.node_id IS NOT NULL
+                 FROM session_artifacts a
+                 LEFT JOIN session_nodes na  ON na.node_id  = a.anchor_node_id
+                 LEFT JOIN session_nodes nss ON nss.node_id = a.source_start_node_id
+                 LEFT JOIN session_nodes nse ON nse.node_id = a.source_end_node_id
+                 WHERE a.session_id = ?1",
+            )
+            .map_err(|error| format!("prepare stale artifact query failed: {error}"))?;
+        let rows = stmt
+            .query_map(rusqlite::params![session_id], |row| {
+                Ok((
+                    row.get::<_, String>(0)?,
+                    row.get::<_, Option<String>>(1)?,
+                    row.get::<_, Option<String>>(2)?,
+                    row.get::<_, Option<String>>(3)?,
+                    row.get::<_, Option<i64>>(4)?,
+                    row.get::<_, Option<i64>>(5)?,
+                    row.get::<_, Option<i64>>(6)?,
+                    row.get::<_, bool>(7)?,
+                    row.get::<_, bool>(8)?,
+                    row.get::<_, bool>(9)?,
+                ))
+            })
+            .map_err(|error| format!("scan stale artifact refs failed: {error}"))?;
+        let mut out = Vec::new();
+        for row in rows {
+            out.push(row.map_err(|error| format!("read stale artifact row failed: {error}"))?);
+        }
+        out
+    };
+
+    let ref_dangles = |node_id: &Option<String>, ti: Option<i64>, exists: bool| -> bool {
+        node_id.is_some() && (!exists || ti.is_some_and(|v| v > new_tail_turn_index))
+    };
+
+    for (
+        artifact_id,
+        anchor_node_id,
+        source_start_node_id,
+        source_end_node_id,
+        anchor_ti,
+        source_start_ti,
+        source_end_ti,
+        anchor_exists,
+        source_start_exists,
+        source_end_exists,
+    ) in &stale_artifact_refs
+    {
+        let anchor_drop = ref_dangles(anchor_node_id, *anchor_ti, *anchor_exists);
+        let source_start_drop =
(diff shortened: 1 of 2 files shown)


Mike-7777777:
Filed Issue #1406 for triage / discussion of the bug independent of this fix.

chumyin:
hi @Mike-7777777, thanks for the careful catch here. i reviewed the full rewrite path, added a small follow-up so artifact head_name cleanup is audited in the same preservation event path, reran the focused local checks, and the refreshed ci on `9948dbec` is green now. nice work on the repro and coverage.

PR #1440, opened 2026-05-03
feat: add OpenTelemetry agent observability with GenAI semantic conventions

cybershang:
## Summary

- Problem: No OTel observability in the agent runtime — unable to collect LLM call traces, token usage, and tool execution telemetry via standard protocols.
- Why it matters: Without standardized trace export, we cannot integrate with Jaeger/Grafana or diagnose LLM latency, failures, and token costs effectively.
- What changed: Added OTLP/HTTP trace export with three span types following OTel GenAI Semantic Conventions (`invoke_agent` / `chat` / `execute_tool`). LLM content capture is off by default and requires opt-in.
- What did not change (scope boundary): Zero function signature changes. Existing `tracing` log pipeline and audit system are untouched. No new kernel-layer spans.

## Linked Issues

- Related # (add corresponding issue if applicable)

## Change Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [x] Daemon / CLI / install
- [x] Providers / routing
- [x] Tools
- [ ] Browser automation
- [ ] Channels / integrations

## Risk Track

## Validation

- [x] `cargo test --workspace --all-features` — 336 passed, 3 failed (pre-existing `onboard_cli` UI width tests, unrelated to this change)
- [x] Jaeger end-to-end verification: OTLP/HTTP → localhost:4318 → spans visible

Commands and evidence:

```text
$ ./scripts/cargo-local-toolchain.sh fmt --all -- --check
(no output — clean)

$ ./scripts/cargo-local-toolchain.sh clippy --workspace --all-targets --all-features -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 17.45s

$ ./scripts/cargo-local-toolchain.sh test --workspace --all-features
test result: 336 passed; 3 failed (pre-existing onboard_cli width tests)

$ OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 OTEL_SERVICE_NAME=loong cargo run --bin loong
→ Jaeger UI shows: invoke_agent loong / chat {model} / execute_tool {tool_name} spans
```

- New environment variables (all optional):
  - `OTEL_EXPORTER_OTLP_ENDPOINT` — OTLP HTTP endpoint (e.g. `http://localhost:4318`)
  - `OTEL_SERVICE_NAME` — service name (defaults to `loong`)
  - `LOONG_OTEL_CAPTURE_CONTENT` — set to `1` to capture LLM input/output message content and tool call payloads
- No UI changes, no behavioral changes to existing functionality

## Failure Recovery

- Fast rollback or disable path: `init_otel()` is a no-op when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset — zero overhead
- Observable failure symptoms: OTel SDK silently degrades if endpoint is unreachable; main flow unaffected; check stderr `tracing` logs for confirmation

## Reviewer Focus

- `crates/daemon/src/observability.rs` — `init_otel()` initialization logic, HTTP vs gRPC transport rationale
- `crates/app/src/chat/chat_surface/app.rs:5796-5869` — `FutureExt::with_context` pattern for cross-`tokio::spawn` context propagation
- `crates/app/src/tools/mod.rs:919-1048` — `Context::with_span` + `attach()` for child span nesting
- `crates/app/src/provider/request_executor.rs:922-1194` — opt-in content capture mechanism and chained `if let` pattern


cybershang:
Jaeger:
msedge_pNgqLHfR1C

LLM Attributes:
msedge_6LONC42cxB

LLM Tool Call:
msedge_UXxm3h4YEE


LLM Messages:
msedge_kfd68GykK3

chumyin (changes requested):
hi @cybershang, thanks for pushing this — i found two blockers before i can sign off:

1. `crates/app/src/tools/mod.rs` now records `tool.payload` for every tool call by default. that payload can contain user content, file contents, or secrets, so this breaks the opt-in capture boundary described in the pr summary. please gate payload capture behind the same explicit content opt-in, or redact it aggressively enough that the default path stays content-safe.

2. `crates/daemon/src/observability.rs` swaps the existing `LOONG_LOG_FORMAT=json` path from the custom formatter to the stock `tracing_subscriber` json formatter. that changes the current log pipeline even though the pr summary says the tracing pipeline is untouched. please either preserve the current json log shape or explicitly scope and validate that operator-visible change.

cybershang:
> hi @cybershang, thanks for pushing this — i found two blockers before i can sign off:
> 
> 1. `crates/app/src/tools/mod.rs` now records `tool.payload` for every tool call by default. that payload can contain user content, file contents, or secrets, so this breaks the opt-in capture boundary described in the pr summary. please gate payload capture behind the same explicit content opt-in, or redact it aggressively enough that the default path stays content-safe.
> 2. `crates/daemon/src/observability.rs` swaps the existing `LOONG_LOG_FORMAT=json` path from the custom formatter to the stock `tracing_subscriber` json formatter. that changes the current log pipeline even though the pr summary says the tracing pipeline is untouched. please either preserve the current json log shape or explicitly scope and validate that operator-visible change.

Thanks for the review. Regarding the two blockers:

For tool.payload capture — I’ve made it gated by the same LOONG_OTEL_CAPTURE_CONTENT environment variable. By default, payload is not captured; only when that opt‑in is enabled will it be recorded. This keeps the default content‑safe behavior.

For observability.rs — I’ve restored the previous JSON formatter (LoongJsonEventFormat / LoongJsonFields) so the log pipeline now matches the original shape again when LOONG_LOG_FORMAT=json. The tracing pipeline remains unchanged as stated in the PR summary.

crates/daemon/src/main.rs
www13059690, line 208: Does "_otel_guard" need to drop in this line?

cybershang: I guess you mean line 172.

crates/daemon/src/main.rs
```diff
@@ -169,6 +169,7 @@ fn check_legacy_home_migration() {
 fn main() {
     let _stdin_guard = StdinGuard;
     init_tracing();
+    let _otel_guard = init_otel();
```
cybershang: @www13059690 You may ask this line.
 The `_otel_guard` holds the `OtelGuard` returned by `init_otel()` for its entire lifetime — it's a pure RAII pattern. 

Without it, the temporary would be dropped at the
 end of the `init_otel();` statement, immediately triggering `SdkTracerProvider::shutdown()` before any real traces are emitted. 

Binding it to `_otel_guard` extends the
 guard's lifetime to match `main()`, ensuring the OTLP exporter stays alive and only flushes on graceful exit.

crates/daemon/src/main.rs
www13059690: I mean, if the app exits at this line, your _otel_guard variable won't be dropped correctly. Is that a problem?

crates/daemon/src/observability.rs
```diff
@@ -332,6 +337,70 @@ fn structured_json_field_value(field_name: &str, value: &str) -> Option<(&'stati
...
+}
+
+pub fn init_otel() -> OtelGuard {
+    if !otel_traces_export_is_enabled() {
+        return OtelGuard { provider: None };
+    }
+
+    let service_name = std::env::var("OTEL_SERVICE_NAME").unwrap_or_else(|_| "loong".to_owned());
+
+    let exporter = match SpanExporter::builder().with_http().build() {
```
www13059690, line 385: Why are we using HTTP rather than HTTPS or TLS? Is that considered secure enough for our use case?

crates/app/src/provider/request_executor.rs
```diff
@@ -931,6 +933,34 @@ where
     PreStatusError: FnMut(&ProviderApiError) -> bool,
 {
     let model_name = runtime.model.to_owned();
+    let provider_name = runtime.provider.kind.as_str().to_owned();
+    let base_url = runtime.provider.base_url.clone();
+
+    let otel_tracer = global::tracer_provider().tracer("loong");
+    let capture_content =
+        std::env::var("LOONG_OTEL_CAPTURE_CONTENT").is_ok_and(|v| v == "1" || v == "true");
```
www13059690, line 941: Why use an env var for this flag rather than a config item?

crates/daemon/src/main.rs
cybershang: Thanks!
 You're right that `std::process::exit()` skips stack destructors, so `OtelGuard::drop()` would never run and buffered spans would be lost. In practice the `exit(2)` paths are early-fatal errors (CLI parse failure, config corruption) that fire before any real trace-worthy work begins, so the practical impact is negligible. 

That said, a `drop`-based RAII guard that can be silently bypassed by an `exit()` is a brittle pattern. 

I'll add an explicit `OtelGuard::shutdown()` call before those exit sites to make it robust regardless of timing.

www13059690 (approved):

chumyin (approved):

crates/daemon/src/observability.rs
www13059690: **cybershang** will fix it in a later PR.

PR #1437, opened 2026-05-02
Refactor tool runtime surfaces and retire browser companion

chumyin:
## Summary

- Problem:
  Loong still carried an overdesigned browser companion / preview lane, model-facing `skills.*` runtime drift, and several stale runtime helpers that no longer matched the direct tool surface.
- Why it matters:
  The extra layers made the runtime harder to reason about, less Pi/Codex-like in behavior, and more fragile around followups, browser-style tasks, managed skills, and operator surfaces.
- What changed:
  - removed the `browser_companion` / `browser_preview` lane from app, daemon, bundled skills, and docs
  - kept the built-in surface centered on bounded `browse`-style page inspection while leaving richer browser automation to `agent-browser`
  - kept skills prompt-first and operator-managed instead of model-facing runtime tools
  - deleted stale runtime modules and helper paths that were no longer on the execution path
  - aligned app, daemon, onboarding, snapshot/restore, and integration tests to the simplified runtime contract
- What did not change (scope boundary):
  - the direct user-facing core tool surface remains centered on `read`, `write`, `edit`, `bash`, `web`, `browser`, and `memory`
  - managed skills still exist as an operator-managed domain
  - no irreversible migration or data-format break was introduced

## Linked Issues

- Closes #1436

## Change Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [x] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [ ] Kernel / policy / approvals
- [ ] Contracts / protocol / spec
- [x] Daemon / CLI / install
- [x] Providers / routing
- [x] Tools
- [x] Browser automation
- [x] Channels / integrations
- [x] ACP / conversation / session runtime
- [x] Memory / context assembly
- [x] Config / migration / onboarding
- [x] Docs / contributor workflow
- [ ] CI / release / workflows

## Risk Track

- [ ] Track A (routine / low-risk)
- [x] Track B (higher-risk / policy-impacting)

If Track B, fill these in:

- Risk notes:
  The main risk is semantic drift across app and daemon after removing the companion lane and deleting stale runtime helpers.
- Rollout / guardrails:
  Full local repo gates were rerun after the final deletion pass, including strict clippy and full all-features tests.
- Rollback path:
  Revert this PR. The refactor is code-structure and runtime-surface cleanup only.

## Validation

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
[...]

PR #1451, opened 2026-05-06
Stabilize IM conversation session routing and explicit reset

### LINKED ISSUE #1450: Stabilize IM conversation session routing and explicit reset semantics
## Problem

IM-backed conversations do not keep a stable local session identity across turns. In practice this can fragment a single chat window into multiple sessions, which breaks the expected mental model for Feishu and other chat integrations.

## Why It Matters

Operators expect one IM conversation window to map to one local Loong session until they explicitly reset it with `/new` or `/reset`.

Without that contract:

- follow-up turns can lose continuity
- previous session ids can become hard to find or resume
- IM behavior diverges from normal CLI session expectations
- each channel risks growing bespoke session handling instead of sharing one runtime rule

## Requested Outcome

Implement a unified IM session-routing model with these guarantees:

- the default session identity for an IM conversation is stable across turns
- `/new` or `/reset` explicitly rotates that conversation onto a fresh local session
- both the previous and the new local sessions remain resumable through the normal CLI/runtime path
- the solution lives in shared IM/session infrastructure rather than as a Feishu-only patch

## Acceptance Signals

- same IM route resolves to the same local session id by default
- explicit reset rotates the route binding to a new local session id
- sender/thread/message context remains available to ingress/reply logic without always fragmenting session identity
- platforms that truly need thread-scoped identity can still opt in explicitly

chumyin:
## Summary

- Problem:
  IM-backed conversations could fragment one chat window into multiple local sessions because session identity was coupled too tightly to message-level context.
- Why it matters:
  Operators expect one IM conversation window to stay on one local Loong session until they explicitly reset it with `/new` or `/reset`, and they expect both old and new sessions to remain resumable.
- What changed:
  Added a shared IM route-binding layer that maps a stable IM route to the current local session, keeps Feishu-style sender/thread/message context in ingress metadata instead of always fragmenting identity, and rotates the binding to a fresh local session on explicit reset commands.
  ACP still supports explicit fine-grained thread/participant routing when a caller intentionally supplies that structured address; this PR only changes the default IM inbound session contract.

## Linked Issues

- Closes #1450
- Related #1437

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [ ] Daemon / CLI / install
- [ ] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [x] Channels / integrations
- [x] Memory / context assembly

## Risk Track

## Validation

- [x] Rust tests match the touched packages / feature surface (for example `cargo test --workspace --locked`, `cargo test --workspace --all-features --locked`, `task verify`, `task verify:changed`, or equivalent exact commands documented below)
- [x] If full workspace / all-features Rust tests were not run locally, explain why and cite CI evidence or a known baseline blocker

Commands and evidence:

```text
cargo fmt --all -- --check
cargo check -p loong-app
cargo test -p loong-app channel::tests:: -- --nocapture
cargo test -p loong-app channel::inbound_turn::tests:: -- --nocapture
cargo test -p loong-app chat::latest_session_selector_tests:: -- --nocapture
cargo test -p loong-app handle_turn_with_runtime_and_address_enforces_account_and_thread_dispatch_scope -- --nocapture
cargo test -p loong-app handle_turn_with_runtime_reuses_shared_acp_session_between_turns -- --nocapture

Result: all passed locally.
Notes:
- I intentionally used targeted crate/test coverage instead of workspace-wide all-features verification because the touched logic is isolated to `loong-app` channel/session runtime surfaces, and the repo memory guidance for this area favored exact changed-surface verification plus crate-level compile checks.
- Added focused regression coverage for:
  - stable IM route -> local session reuse
  - explicit `/new` and `/reset` binding rotation
  - reopening both the current and prior IM local session ids through the normal CLI runtime path
  - preserving Telegram thread/topic-scoped identity where the adapter explicitly opts in
```

- IM conversations now default to one stable local session per conversation window instead of fragmenting around message-level Feishu context.
- `/new` and `/reset` now rotate that IM conversation onto a fresh local session while preserving the previous session for later reopen/resume.
- Feishu keeps sender/thread/message context for ingress and reply behavior without treating every message/thread detail as a new default session.

## Failure Recovery

  Revert commit `d658a06c5f9663a12d33617838f22d4d9b506022`.
  - IM follow-up turns unexpectedly opening brand-new local sessions without `/new` or `/reset`
  - Telegram topic/thread conversations collapsing into parent-chat session identity
  - Explicitly reopened IM local session ids failing to load through the CLI/runtime path

## Reviewer Focus

- `crates/app/src/channel/inbound_turn.rs`: shared IM route-binding creation, reset rotation, and ACP cleanup handoff.
- `crates/app/src/channel/core/types.rs`: default identity vs opt-in participant/thread identity split.
- `crates/app/src/channel/telegram.rs`: explicit topic/thread identity opt-in for Telegram forum-style windows.
- `crates/app/src/session/repository.rs` and `crates/app/src/memory/sqlite.rs`: persistent `session_route_bindings` storage and lookup semantics.

### DIFF
--- a/crates/app/src/channel/core/types.rs
+++ b/crates/app/src/channel/core/types.rs
@@ -153,6 +153,10 @@ pub struct ChannelSession {
     pub conversation_id: String,
     pub participant_id: Option<String>,
     pub thread_id: Option<String>,
+    #[serde(default)]
+    pub identity_participant_scoped: bool,
+    #[serde(default)]
+    pub identity_thread_scoped: bool,
 }
 
 impl ChannelSession {
@@ -164,6 +168,8 @@ impl ChannelSession {
             conversation_id: conversation_id.into(),
             participant_id: None,
             thread_id: None,
+            identity_participant_scoped: false,
+            identity_thread_scoped: false,
         }
     }
 
@@ -179,6 +185,8 @@ impl ChannelSession {
             conversation_id: conversation_id.into(),
             participant_id: None,
             thread_id: None,
+            identity_participant_scoped: false,
+            identity_thread_scoped: false,
         }
     }
 
@@ -194,6 +202,8 @@ impl ChannelSession {
             conversation_id: conversation_id.into(),
             participant_id: None,
             thread_id: Some(thread_id.into()),
+            identity_participant_scoped: true,
+            identity_thread_scoped: true,
         }
     }
 
@@ -210,6 +220,8 @@ impl ChannelSession {
             conversation_id: conversation_id.into(),
             participant_id: None,
             thread_id: Some(thread_id.into()),
+            identity_participant_scoped: true,
+            identity_thread_scoped: true,
         }
     }
 
@@ -225,6 +237,16 @@ impl ChannelSession {
 
     pub fn with_thread_id(mut self, thread_id: impl Into<String>) -> Self {
         self.thread_id = Some(thread_id.into());
+        self
+    }
+
+    pub fn with_identity_participant_scoped(mut self, scoped: bool) -> Self {
+        self.identity_participant_scoped = scoped;
+        self
+    }
+
+    pub fn with_identity_thread_scoped(mut self, scoped: bool) -> Self {
+        self.identity_thread_scoped = scoped;
         self
     }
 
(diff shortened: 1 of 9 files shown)


chumyin:
merged after a final pass on the im route-binding flow. i also tightened reset-generated local session ids so rapid explicit resets cannot collide, refreshed the pr description to remove local machine paths, and re-ran the focused local checks before waiting for the full ci suite to go green. thanks for pushing this through.

PR #1505, opened 2026-05-29
fix: run OTLP batch processor on Tokio instead of futures_executor

cybershang:
## Summary

- **Problem:** `BatchSpanProcessor`'s background thread used `futures_executor::block_on`, which lacks a Tokio runtime. When `reqwest` tried DNS resolution via `hyper-util`'s `GaiResolver`, it called `tokio::runtime::Handle::current()` and panicked because no Tokio reactor was present. Additionally, CA cert handling used `panic!()` calls that would crash the daemon on invalid cert configurations.
- **Why it matters:** The observability feature is unusable in any deployment that actually sends traces — the daemon panics on startup whenever OTLP export is active. This blocks the entire OTel observability feature from working in production and development alike.
- **What changed:**
  - Switched from `BatchSpanProcessor` (futures_executor) to `BatchSpanProcessor` with `runtime::Tokio` from `opentelemetry-sdk::trace::span_processor_with_async_runtime`
  - Moved `init_otel()` inside `runtime.block_on()` so `tokio::spawn` has an active Tokio context
  - Replaced `panic!()` calls in CA certificate handling with graceful `Err` returns
  - Updated `Cargo.toml` dependency to pull in the async runtime module
- **What did not change (scope boundary):** No changes to the OTel collector config, no changes to trace data format or sampling, no changes to Jaeger or any other observability backend. Pure runtime fix — the OTel agent feature's API surface and configuration schema are untouched.

## Linked Issues

- Closes # (none)
- Related # (none)

## Change Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Security hardening
- [ ] CI / workflow / release

## Touched Areas

- [x] Daemon / CLI / install
- [ ] Providers / routing
- [ ] Tools
- [ ] Browser automation
- [ ] Channels / integrations

## Risk Track

If Track B, fill these in:

- **Risk notes:** Replaces the async runtime for the OTLP batch processor background thread. While `runtime::Tokio` is the canonical choice for a Tokio-based application, any regression would manifest as trace data not being exported (silent data loss) or a daemon crash on startup.
- **Rollout / guardrails:** The change is isolated to the `init_otel()` function and its call site in `main.rs`. The original `BatchSpanProcessor` path is fully replaced — no fallback path is provided, so testing the OTLP export path end-to-end after deploy is essential.
- **Rollback path:** Revert to the previous `BatchSpanProcessor` with `futures_executor`. The old code path is known to panic on DNS resolution, so rollback restores the broken state — this fix should only be rolled forward.

## Validation

- [ ] Rust tests match the touched packages / feature surface
- [x] Full workspace / all-features Rust tests were not run locally

Commands and evidence:

```text
# Format check passed — no format issues in the changed files.
cargo fmt --all -- --check

# Clippy and full test suite not run locally; this is a single-function
# change in a feature-gated observability module. Recommend CI validation
# before merge, specifically:
cargo test --workspace --all-features --locked   # full regression
```

- **Before:** Daemon panics at startup if OTLP trace export is enabled, with a `tokio::runtime::Handle::current()` panic inside `reqwest` DNS resolution. CA certificate misconfiguration also causes an unconditional panic.
- **After:** OTLP batch processing runs on the Tokio runtime correctly. CA cert errors return a graceful error message instead of crashing the daemon.

## Failure Recovery

- **Fast rollback or disable path:** Remove the `otel` feature flag or revert the commit. Note that rolling back re-exposes the panic-on-startup bug for anyone using OTLP export.
- **Observable failure symptoms reviewers should watch for:** Daemon fails to start with OTel enabled (regression), or trace data silently not being exported (batch processor not flushing).

## Reviewer Focus

- `crates/daemon/src/observability.rs` — the core fix: verify the `BatchSpanProcessor` is correctly constructed with `runtime::Tokio`, and that CA error handling returns `Err` instead of panicking.
- `crates/daemon/src/main.rs` — verify `init_otel()` is called inside `runtime.block_on()` so Tokio is active.
- `Cargo.toml` — confirm the `opentelemetry-sdk` dependency includes the async runtime feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ tracing = "0.1"
 tracing-log = "0.2"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt", "json"] }
 opentelemetry = { version = "0.29", features = ["trace", "futures"] }
-opentelemetry_sdk = { version = "0.29", features = ["rt-tokio", "trace"] }
+opentelemetry_sdk = { version = "0.29", features = ["rt-tokio", "trace", "experimental_async_runtime", "experimental_trace_batch_span_processor_with_async_runtime"] }
 opentelemetry-otlp = { version = "0.29", features = ["trace", "http-proto"] }
 zip = { version = "8", default-features = false, features = ["deflate"] }
 
--- a/crates/daemon/src/main.rs
+++ b/crates/daemon/src/main.rs
@@ -169,7 +169,6 @@ fn check_legacy_home_migration() {
 fn main() {
     let _stdin_guard = StdinGuard;
     init_tracing();
-    let mut _otel_guard = init_otel();
     mvp::config::set_active_cli_command_name(mvp::config::detect_invoked_cli_command_name());
     loong_daemon::make_env_compatible();
     check_legacy_home_migration();
@@ -189,8 +188,12 @@ fn main() {
         command = %redacted_command,
         "resolved CLI command"
     );
-    let result = build_daemon_runtime(&command)
-        .and_then(|runtime| runtime.block_on(run_command(command, invoked_as_default_entry)));
+    let result = build_daemon_runtime(&command).and_then(|runtime| {
+        runtime.block_on(async {
+            let _otel_guard = init_otel();
+            run_command(command, invoked_as_default_entry).await
+        })
+    });
     if let Err(error) = result {
         let error_code = error_code(error.as_str());
         tracing::error!(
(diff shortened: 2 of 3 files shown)


www13059690 (approved):
LGTM. Moving OTEL initialization into the Tokio runtime and switching the batch span processor from the old dedicated thread path to a Tokio-task-based path makes the execution model much cleaner for this daemon, especially since the exporter is using async HTTP underneath. I also like the client-side certificate support: allowing a custom CA to be injected into the OTLP reqwest client via OTEL_CA_CERT_FILE makes the tracing pipeline more practical for internal or self-hosted collectors with non-default trust roots. Overall, this change improves both runtime alignment and deployability.

PR #1519, opened 2026-06-13
Optimize tui

InuDial:
From #1516


www13059690 (approved):
LGTM. I reviewed the updated diff and the remaining behavior changes look intentional: the test_support split, the tool-shell gating changes, and the TUI module split all
  compile cleanly under the checked targets. I’m comfortable approving this for merge.

PR #1524, opened 2026-06-16
refactor: name tui surface layout height parts

hugo1120:
## Summary

  This PR is a small clean code/refactor pass for chat surface layout height calculation.

  The goal is to make the TUI surface rendering layout easier to read by replacing unnamed height
  literals with named constants and small helper functions.

  ## What Changed

  - Replaced repeated layout height literals with named constants:
    - `COMPOSER_SEPARATOR_HEIGHT`
    - `PALETTE_SEPARATOR_HEIGHT`
    - `FOOTER_SEPARATOR_HEIGHT`
    - `FOOTER_HEIGHT`
  - Extracted bottom layout height calculation into small pure helpers:
    - `palette_band_height`
    - `bottom_band_height`
  - Renamed the `RenderSectionHeights` field `comsposer` to `composer`.
  - Removed the old `TODO: translate magic numbers` comment by making the height parts explicit.
  - Added a focused regression test for palette-visible vs palette-hidden bottom band height
  calculation.

  ## Why

  `compute_render_section_heights` previously used several bare `1` values in one expression. Those
  values were correct, but the intent was not obvious.

  This refactor gives each fixed row a name and keeps the total reserved bottom height in one helper.

  ## Behavior / Compatibility

  This is intended to be behavior-preserving.

  It does not change render section order, transcript height behavior, composer height behavior,
  palette visibility behavior, footer rendering, or public chat surface APIs.

  ## Tests

  ```text
  rustfmt --check crates/app/src/chat/chat_surface/app/surface.rs
  git diff --check
  cargo test -p loong-app bottom_band_height_includes_palette_separator_only_when_palette_is_visible
  -j 1
  cargo test -p loong-app footer -j 1
  cargo clippy -p loong-app --all-targets --all-features -- -D warnings

  ## Notes for Reviewers

  The main review point is that bottom_band_height matches the old layout arithmetic while making each
  reserved row explicit.

### DIFF
--- a/crates/app/src/chat/chat_surface/app/surface.rs
+++ b/crates/app/src/chat/chat_surface/app/surface.rs
@@ -5,7 +5,7 @@ pub struct RenderSectionHeights {
     spacer: u16,
     pending: u16,
     composer_separator: u16,
-    comsposer: u16,
+    composer: u16,
     palette_separator: u16,
     palette: u16,
     footer_separator: u16,
@@ -20,7 +20,7 @@ impl RenderSectionHeights {
             self.spacer,
             self.pending,
             self.composer_separator,
-            self.comsposer,
+            self.composer,
             self.palette_separator,
             self.palette,
             self.footer_separator,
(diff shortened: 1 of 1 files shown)


www13059690 (approved):
LGTM, approved.

This PR is well scoped and only refactors the TUI surface layout height calculation in `surface.rs`. Replacing the unnamed height literals with named constants and extracting `palette_band_height` / `bottom_band_height` makes the layout arithmetic easier to read.

I checked the old and new height calculations, including both palette-hidden and palette-visible cases, and did not find any behavior change.

